PR #127, opened 2013-09-03
private[scalatest] 'contain AMatcher/AnMatcher'

cheeseng:
Mark methods that enable 'contain AMatcher/AnMatcher' as private[scalatest], should re-enable them back when it is completed like others contain xxx using type classes.

### DIFF
--- a/src/main/scala/org/scalatest/words/ContainWord.scala
+++ b/src/main/scala/org/scalatest/words/ContainWord.scala
@@ -163,7 +163,7 @@ final class ContainWord {
    *                                ^
    * </pre>
    */
-  def a[T](aMatcher: AMatcher[T]): Matcher[GenTraversable[T]] = 
+  private[scalatest] def a[T](aMatcher: AMatcher[T]): Matcher[GenTraversable[T]] =
     new Matcher[GenTraversable[T]] {
       def apply(left: GenTraversable[T]): MatchResult = {
         val matched = left.find(aMatcher(_).matches)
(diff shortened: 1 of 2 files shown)


rtyley:
Given that these methods are private-ish, what's best way to convert a `Matcher[T]` into a `Matcher[Traversable[T]` (ie "contains an element that matches matcher `X`"?

bvenners:
@rtyley You can use an Inspector:

forAtLeast(1, xs) { x =>
  x should be > 7
}

Or inspector shorthands:

atLeast (1, xs) should be > 7

rtyley:
Thanks @bvenners - while I can see that `forAtLeast()` works in your example, unfortunately what I would _like_ is to convert a `Matcher[T]` into a `Matcher[Traversable[T]` (which `forAtLeast()` does _not_ do), to allow me to pass it around as a parameter to functions.

Specifically, this is the bit of code I'm looking at right now, trying to convert from specs2 to scalatest. It's attempting to test that a Git repository a) contains bad history before the BFG runs, and b) no longer contains bad history after the BFG runs:

```
ensureRemovalOf(commitHistory(haveCommitWhereObjectIds(contain(be_==(abbrId("294f")))).atLeastOnce)) {
  run("--strip-blobs-bigger-than 1K")
}
```
https://github.com/rtyley/bfg-repo-cleaner/blob/3fb584a9/bfg/src/test/scala/com/madgag/git/bfg/cli/MainSpec.scala#L39

The `ensureRemovalOf()` method looks like this:

```
def ensureRemovalOf[T](dirtMatchers: Matcher[Repository]*)(block: => T) = {
  repo must (dirtMatchers.reduce(_ and _))
  block
  repo must dirtMatchers.map(not(_)).reduce(_ and _)
}
```
https://github.com/rtyley/bfg-repo-cleaner/blob/3fb584a/bfg-test/src/main/scala/com/madgag/git/bfg/test/unpackedRepo.scala#L91-L97

...so I'm passing `Matcher`s around, and I want to be able to match on the commits of repository history containing a bad file (ie a `Matcher[Seq[Commit]` ) - and then later check that the matcher no longer matches the updated history.

As ScalaTest's `forAtLeast()` inspector doesn't return a `Matcher[Traversable[T]`, I can't use it in a simple translation of my existing test code.

PR #365, opened 2014-08-07
Fixed typo

tkawachi:


rfries:
Hi, and thanks for contributing to ScalaTest!  We'd like to include this pull request in an upcoming release, but we will need to get a Contributor License Agreement (CLA) on file for you first (this only needs to be done once).

if the contributor is a corporation, there is a different version:

This can be returned to cla@artima.com or to the address or fax number listed in the document.

Also, since this would be included in an upcoming 2.3.x release, could you resubmit against the current 2.3.x branch?

Thanks again for your contribution.

plinlor:
@cla-bot[bot] check

PR #384, opened 2014-09-10
Adds a test which demonstrates a bug in combining OIPT and BeforeAndAfter

mosesn:
I don't actually understand what the problem is, but the before block is being executed more than once per test.

@bvenners any ideas?

### DIFF
--- a/src/test/scala/org/scalatest/OneInstancePerTestSpec.scala
+++ b/src/test/scala/org/scalatest/OneInstancePerTestSpec.scala
@@ -208,5 +208,26 @@ class OneInstancePerTestSpec extends FunSpec {
         }
       }
     }
+
+    it("should only run before / after once per test") {
+      @volatile var x = 0
+      class FooSuite extends FunSuite with OneInstancePerTest with BeforeAndAfter {
+        before {
+          println("WHATEVER")
+          x += 1
+        }
+
+        test("x is 2") {
+          assert(x === 2)
+        }
+
+        override def newInstance = new FooSuite
+      }
+
+      val rep = new EventRecordingReporter
+      val outer = new FooSuite
+      outer.run(None, Args(rep))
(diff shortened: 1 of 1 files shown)


rfries:
So I think what is happening here is that the outer.run() call is actually creating two instances of the test case.  Since the "x" counter was in the test case instance, both were incrementing x, which produces undefined results (on my machine, x was always 0, not 2).

Although two instances are being created, the test case is only invoked once.  This can be seen more clearly with the following version of the test:

```scala
      val count = new AtomicInteger(0)
      class FooSuite extends FunSuite with OneInstancePerTest with BeforeAndAfter {
        val id = count.incrementAndGet()
        var x = 0
        before {
          println(s"$id: before")
          x += 1
        }
        test("x is 1") {
          println(s"$id: test...")
          assert(x === 1)
        }
        after {
          println(s"$id: after")
        }
        override def newInstance = new FooSuite
      }
      val rep = new EventRecordingReporter
      val outer = new FooSuite
      println("calling run...")
      outer.run(None, Args(rep))
      assert(rep.testSucceededEventsReceived.size === 1)
```

which produces the following output

```
calling run...
1: before
2: before
2: test...
2: after
1: after
```

ashawley:
Fixed in #597?

```scala
it("should only run before / after once per test") {
  @volatile var x = 0
  class FooSuite extends FunSuite with OneInstancePerTest with BeforeAndAfter {
    before {
      println("WHATEVER")
      x += 1
    }

    test("x is 1") {
      assert(x === 1) // ✓
    }

    override def newInstance = new FooSuite
  }

  val rep = new EventRecordingReporter
  val outer = new FooSuite
  outer.run(None, Args(rep))
  assert(rep.testSucceededEventsReceived.size === 1)
}
```

plinlor:
@cla-bot[bot] check

PR #451, opened 2014-11-26
Add status to Shell

thetrav:
This allows programmatic invocation of Shell.run and interrogation of results

### DIFF
--- a/src/main/scala/org/scalatest/Shell.scala
+++ b/src/main/scala/org/scalatest/Shell.scala
@@ -377,7 +377,7 @@ sealed trait Shell {
    * this method is invoked.
    * </p>
    */
-  def run(suite: Suite, testName: String = null, configMap: ConfigMap = ConfigMap.empty): Unit
+  def run(suite: Suite, testName: String = null, configMap: ConfigMap = ConfigMap.empty): Status
 }
 
 // parameters were private, but after pulling out the trait so I don't import copy() as part
@@ -401,7 +401,7 @@ private[scalatest] final case class ShellImpl(
   lazy val nostacks: Shell = copy(shortstacksPassed = false, fullstacksPassed = false)
   lazy val nostats: Shell = copy(statsPassed = false)
 
-  def run(suite: Suite, testName: String = null, configMap: ConfigMap = ConfigMap.empty) = {
+  def run(suite: Suite, testName: String = null, configMap: ConfigMap = ConfigMap.empty): Status = {
     suite.execute(testName, configMap, colorPassed, durationsPassed, shortstacksPassed, fullstacksPassed, statsPassed)
   }
 }


bvenners:
Hi Travis,

I like the idea in principle to have a method that returns Status, but it will add a line to the REPL that isn't really needed. I'm thinking it might be better to have a different method intended for programmatic access, but I'd like to better understand what programmatic access might be desired. Can you let me know more about your use case? What are you getting by programmatically interacting with Shell that you can't get by programmatically interacting with Suite directly. (Note there are execute methods on Suite that have a simpler API than the run method on Suite.)

Thanks.

Bill

joune:
Hi Bill,

I'm jumping in because I also have a similar need to programmatically execute test suites. 

From the Suite scaladoc it looks like I have an option of calling 'execute' or 'run' on a Suite, where execute will give me a standard output reporter but no status and run will give me a status but no reporter.. I checked the source code for the Runner but all implementations for the options (stdOutReporter, junitReporter..) are private

Something like:

    (new Test run(None, new Args(new Reporter { def apply(e:events.Event) = println(e) }))).succeeds

is close to what i need, but I have to re-implement the reporters.

The context is to run tests in an OSGi container, but regardless of the details, the need is to execute tests on a Suite that is programmatically instantiated and be able to reuse the same options supported by tools.Runner. 

Even without OSGi, the same need would arise with conditional mixins, for instance:

    val test  = if (headless) new MyTest with HtmlUnitDriver else new MyTest with FirefoxDriver

This kind of thing cannot be passed as a String the Runner would instantiate. 

wdyt?

bvenners:
Hi All,

Unrelated to this ticket a few days ago I modified the execute methods of Suite such that they will not return until execution of the suite has completed:

https://github.com/scalatest/scalatest/commit/5501398ee0fdef5faf7a2ebf6754b03baed3acc2

I'm curious if this will address Arjun's concern at least, and possibly also thetrav's. Is the reason you wanted the status that you wanted to block and wait until it was completed? Or did you want to know if it succeeded or failed perhaps as well? Something else?

I probably would not be in favor of changing the result type of Shell.run or Suite.execute from Unit to Status, but could be persuaded to add a different method that does that. But first I'd like to understand the use cases.

Thanks.

Bill

joune:
Thanks Bill, but, not quite, I'm afraid.

I had not even realized that the execute method was "asynchronous"; but returning Unit and without a callback of some kind, this seems like a welcome change.

In addition to this, I think the need can be decomposed as such:

* programmatically run a test that I instantiated myself (as such, it's covered by execute)
* have the means to know wether it succeeded (Status or at least a boolean)
* have the means to know when it's over (either synchronously as you did, or returning a Future[Status]) so that I can chain another action (another suite or a System.exit)
* be able to reuse existing reporters (i would typically need stdout and junit)

It's a long list but in essence I just mean a "programmable" version of tools.Runner. 
Maybe an easy path could be an additional method like tools.Runner.run that would accept a Seq[Suite] "in place of" (or in addition to) the "-s" parameter, and with all the same other arguments?

bvenners:
Perhaps we could add a runSuites method to Runner that takes Seq[Suite] as you suggest. Probably we'd disallow -s, but I'll have to look. And runSuites could return a Status. (Or it could also be an overloaded run method.) I think it does belong in Runner more than Suite.

Bill

plinlor:
@cla-bot[bot] check

PR #465, opened 2014-12-16
Unprivated the StackDepth code so that those writing matcher libraries can use them

bwmcadams:
& leverage them without monkey patching


plinlor:
@cla-bot[bot] check

PR #475, opened 2014-12-27
Add "exists" to GeneratorDrivenPropertyCheckers

flicken:
Two things to note about this, based on ScalaCheck's Prop.exists method implementation.

  1. Like ScalaCheck, only single-arity functions are supported (e.g. exists(...) { a: String => assert(a.length == 1)  }, but not { (a: String, b: String) => ...}
  2. MinSuccessful and MaxDiscardedFactor config settings appear to *not* be supported by Prop.exists.   That is, checking Prop.exists with different maxDiscardedRatio values does not appear to affect the result.   I've disabled directly passing minSuccessful, etc property values, but confusion could arise because the user can still set the minSuccessful, etc properties on a PropertyCheckConfiguration instance.

### DIFF
--- a/project/GenGen.scala
+++ b/project/GenGen.scala
@@ -17,6 +17,8 @@ import java.io.File
 import java.io.FileWriter
 import java.io.BufferedWriter
 import java.util.Calendar
+import org.antlr.stringtemplate.StringTemplate
+
 import scala.collection.JavaConversions._
 
 object GenGen {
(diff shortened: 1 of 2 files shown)


bvenners:
We decided long ago not to actually go in this direction. Although the code all worked, it added complexity that outweighed the simplicity of the consistency it added.

PR #494, opened 2015-01-29
Basic scaladocs for EitherSugar and TrySugar. [SCALATEST-485]

rfries:
These are the basic scaladocs; some examples similar to those in OptionSugar will follow.


rfries:
Made against the wrong branch, should be 2.2.x....

PR #509, opened 2015-02-11
Travis Maven Cache.

cheeseng:
Added cache to .travis.yml to cache maven dir.  This should improve the build time and reduce a factor of failure due to download error of dependency artifacts.


cheeseng:
Going to submit through 2.3.x instead.

PR #516, opened 2015-02-16
SuiteAborted Fix

cheeseng:
- Changed code to fire SuiteAborted when error occurs during suite construction, and rethrow if it is an fatal error (use NonFatal to determine).
- Make Suite.execute to propagate fatal error.


alexeyts:
Hi,
We're experiencing the issue mentioned here:
https://issues.jenkins-ci.org/browse/JENKINS-22703

It seems this PR should solve it.
What released scalatest version can we use, which includes this PR?
We're using scala 2.10

alexeyts:
We tried release 3.0.0-RC4, which should contain the fix, but unfortunately it doesn't solve our issue.

thomg:
Likewise, we're still seeing the same behavior using the new 3.0.0 release.

robth:
Same here; ScalaTest 3.0.0 / Jenkins 1.625.1.

PR #523, opened 2015-03-01
Enhancenment WebSocketTests

janesser:
In need of testing scalatra atmosphere I did not find a dedicated testing framework for async websockets.

So this pull-request contains a testing based on jetty-websocket-client mixed as a Suite. See an easy-to-understand example with an websocket-echo-server.
(I had to upgrade the version of jetty contained, and the existing JettySpec is fixed / should be still working.)

Don't hesitate to comment.

Kind regards, Jan


janesser:
sbt test or testQuick run until a certain point and than hangs at different points.

I hope Travis CI will check for that correctly.

janesser:
outdated

PR #532, opened 2015-03-20
Scala.js Support M1

cheeseng:
Milestone 1 for scala.js support, first ScalaTest Suite (FunSuite) and assertion test runs via scala.js.


cheeseng:
These changes are already in 2.3.x branch.

PR #540, opened 2015-04-05
Fixes #472 - changes EasyMockSugar#whenExecuting to take in a T, rather ...

### LINKED ISSUE #472: EasyMockSugar, whenExecuting MockObjects form should accept a function that returns Any
Just a very minor annoyance... but the form of whenExecuting that takes an implicit MockObject parameter, like this:

  // http://doc.scalatest.org/2.2.1/index.html#org.scalatest.mock.EasyMockSugar
  def whenExecuting(fun: => Unit)(implicit mocks: MockObjects)

It would be nice if this took fun: => Any instead, because if the last statement in your function (i.e., the test code) doesn't return unit, this form of whenExecuting is not chosen by the compiler, and it won't compile.

The use case I'm seeing right now is where I'm testing some akka code with TestKit and the last statement inside my whenExecuting block is sometimes a test probe's "expectMsg" or similar that returns a value instead of

jleskovar:
...than Unit


plinlor:
@cla-bot[bot] check

PR #559, opened 2015-04-21
Move unzip/unzip3 from EquaSet/SortedEquaSet to LazyBag/LazySeq.

rfries:
This moves the unzip-ish methods from EquaSet and SortedEquaSet into LazyBag and LazySeq.

Note that the tests verify correct semantics, but do not test for laziness.


rfries:
Ugh.  Wrong branch, will re-open....

PR #568, opened 2015-04-25
Change various Gen* return types in EquaSet/SortedEquaSet

rfries:
also removed vestiges of zip/unzip/scan from concrete implementations of EquaSet/SortedEquaSet, and added some additional tests.


plinlor:
@cla-bot[bot] check

PR #597, opened 2015-05-30
Fix issue with using BeforeAndAfter with OneInstancePerTest

rfries:
When BeforeAndAfter is mixed in prior to OneInstancePerTest, it will cause before() and after() to be called in the "outer" instance of the test, even though the test case is not run in that instance.

This is because args.runTestInNewInstance is not being checked, as it is in the implementation of BeforeAndAfterAll (which does not have this issue).

Thanks to @mosesn for pointing out this issue in https://github.com/scalatest/scalatest/pull/384


rfries:
Merged in recent changes from 2.3.x.

PR #598, opened 2015-05-30
Pass-through return value from supplied function in withScreenshot

rfries:


rfries:
Added tests for withScreenshot, and merged in recent changes from 2.3.x.

PR #600, opened 2015-06-01
Updated ConfigMapWrapperSuite to use a ConfigMap constructor

jlawrienyt:
The documentation for ConfigMapWrapperSuite indicates that the test suite should have a single argument in its constructor of type ConfigMap.  However, it was in fact looking for a constructor of type Map[_, _].  I've updated the code to look for a ConfigMap constructor and if not found, to fall back to using the Map[_, _] constructor if present.


plinlor:
@cla-bot[bot] check

PR #628, opened 2015-06-25
requireNonNull should take AnyRef

cheeseng:
Make requireNonNull to take AnyRef instead of Any.


cheeseng:
Decided not to do this anymore.

PR #640, opened 2015-07-13
OSGi Import Package header change

rcostine:
This is the OSGi "import package" change that adds resolution:=optional option to the manifest file for packages that optionally need to be there at a particular release.


plinlor:
@cla-bot[bot] check

PR #659, opened 2015-08-14
Fix isNestedMethod() detection to include a corner case

sralmai:
Nested methods can end with a number including 0. For example,
consider a block like

    class ExampleSpec extends Spec {
      object `An example` {
          def `test 1` {
            def helper() = ...
          }

          def `test 2` {
            def helper() = ...
          }

          ...

          def `test 10` {
            def helper() = ...
          }
        }
      }

The `helper` methods will be turned into something like:

    `ExampleSpec$An$u0020example$$helper$1`
    `ExampleSpec$An$u0020example$$helper$2`
    ...
    `ExampleSpec$An$u0020example$$helper$10`

But that tenth helper method will currently be counted as a test and executed.
(This extra execution is most obvious when the helper method throws an exception.)

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/Spec.scala
+++ b/scalatest/src/main/scala/org/scalatest/Spec.scala
@@ -1420,7 +1420,7 @@ private[scalatest] object Spec {
     
     val isOuterMethod = m.getName.endsWith("$$outer")
     
-    val isNestedMethod = m.getName.matches(".+\\$\\$.+\\$[1-9]+")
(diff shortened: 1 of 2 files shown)


bvenners:
Ah, that's a good one. Sorry about that. I never thought to write a test for that. We'll include this in the next release.

sralmai:
Wow that was a fast response @bvenners ! Thanks for writing awesome software.

plinlor:
@cla-bot[bot] check

PR #673, opened 2015-09-03
Fix Scalacheck Fork Problem

cheeseng:
Fixed scalacheck problem when used in SBT fork mode.

Note: This fix should be cherry-pick into 3.0.x branch.

### DIFF
--- a/project/scalatest.scala
+++ b/project/scalatest.scala
@@ -10,7 +10,7 @@ object ScalatestBuild extends Build {
 
   val buildScalaVersion = "2.11.2"
 
-  val releaseVersion = "2.2.5"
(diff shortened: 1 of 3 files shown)


bvenners:
This fix is released in ScalaTest 2.2.6-SNAP1.

mushtaq:
@cheeseng Can you please add to the next RC of 3.x? I just downgraded to 2.2.6 because our tests must run in the forked mode.

cheeseng:
@mushtaq this fix seems to be already in 3.0.x branch, do you still encounter similar problem with 3.0.0-RC3?  For 3.x version it is using ScalaCheck 1.13.1 though, I wonder if it is related to ScalaCheck version difference?

PR #687, opened 2015-09-08
Add optional OSGI imports to scalatest.scala project file

rcostine:
This adds the resolution:=optional qualifier to imported packages that should be listed as optional in the artifact bundle META-INF/MANIFEST.MF file. If this is not done, you need to add maven (or sbt) dependencies for optional packages in artifacts that may never be used.


cheeseng:
These changes are already in 3.0.x.

Issue #635, opened 2015-07-05
java.lang.NoClassDefFoundError: org/scalactic/exceptions/NullArgumentException


antonkulaga:
when I switch my project ( https://github.com/denigma/akka-http-extensions ) from 3.0.0-M3  to scalatest %%% 3.0.0-M5 (same problem resides in M4) 
```scala
[error] Referring to non-existent class Lorg_scalactic_exceptions_NullArgumentException
[error]   called from Lorg_scalactic_Requirements$RequirementsHelper.macroRequireNonNull__AT__AO__V
[error]   called from Lorg_scalatest_DynaTags.init___sci_Map__sci_Map
[error]   called from Lorg_scalatest_tools_TaskRunner.execute__Lsbt_testing_EventHandler__ALsbt_testing_Logger__ALsbt_testing_Task
[error]   called from Lorg_scalatest_tools_TaskRunner.execute__Lsbt_testing_EventHandler__ALsbt_testing_Logger__F1__V
[error]   called from Lorg_scalajs_testinterface_internal_Slave.$$anonfun$9__p2__Lsbt_testing_Task__Lorg_scalajs_testinterface_internal_Slave$RemoteEventHandler__sjs_js_Array__V
[error]   called from Lorg_scalajs_testinterface_internal_Slave.execute__p2__sjs_js_Dynamic__V
[error]   called from Lorg_scalajs_testinterface_internal_Slave.$$anonfun$1__p2__T__F0__V
[error]   called from Lorg_scalajs_testinterface_internal_Slave.handleMsgImpl__T__F0__V
[error]   called from Lorg_scalajs_testinterface_internal_BridgeBase.handleMsg__p1__T__V
[error]   called from Lorg_scalajs_testinterface_internal_BridgeBase.$$anonfun$1__p1__T__V
[error]   called from Lorg_scalajs_testinterface_internal_BridgeBase.init__V
[error]   called from Lorg_scalajs_testinterface_internal_BridgeBase.$$js$exported$meth$init__O
[error]   called from Lorg_scalajs_testinterface_internal_BridgeBase.init
[error]   exported to JavaScript with @JSExport
[error] involving instantiated classes:
[error]   Lorg_scalactic_Requirements$RequirementsHelper
[error]   Lorg_scalatest_tools_TaskRunner
[error]     (already seen, not repeating call stack)
[error]   Lorg_scalajs_testinterface_internal_Slave
[error]     (already seen, not repeating call stack)
[...]

PR #689, opened 2015-09-08
Add org.scalactic.exceptions package to the scalatest-all jar.

rcostine:

https://github.com/scalatest/scalatest/issues/635


plinlor:
@cla-bot[bot] check

Issue #547, opened 2015-04-17
Can't use -z with test name starting with `exclude`


alexeyr:
I have a test called `exclude_nested_map_with_scalar_from_flattening`. When trying to run it with `-z` in ScalaTest 2.2.4, I get
```
> testOnly scalan.FlatteningTests -- -z exclude_nested_map_with_scalar_from_flattening
WARNING: Argument 'exclude' has been deprecated and will be removed in a future version of ScalaTest.  Please use -l instead.
[trace] Stack trace suppressed: run last enterprise-edition/test:testOnly for the full output.
[error] (enterprise-edition/test:testOnly) java.lang.IllegalArgumentException: Invalid configuration, example valid configuration: exclude(a, b, c)
[error] Total time: 0 s, completed Apr 17, 2015 3:51:22 PM
```

bvenners:
Sorry, that's a good one! I believe we removed those deprecation warnings entirely from 2.3.x, but we'll double check.

PR #692, opened 2015-09-10
2.2: FriendlyParamsTranslator Fix

cheeseng:
Fixed can't use -z with test name starting with 'exclude' problem as reported here: https://github.com/scalatest/scalatest/issues/547


cheeseng:
No applicable anymore, there's another PR to remove FriendlyParamsTranslator from 3.0.x instead: 

https://github.com/scalatest/scalatest/pull/693

PR #698, opened 2015-09-12
2.2: Prettifier ArrayOps Support

cheeseng:
Enhanced Prettifier to support ArrayOps.


cheeseng:
We'll fix this in 3.0.x.

PR #700, opened 2015-09-14
Generators for efficiently generating sets of ordered pairs.

rfries:
Contains generators for basic numeric types (Long, Int, Char, Byte, Double, and Float) for producing pairs (a, b) such that a < b or a <= b.


plinlor:
@cla-bot[bot] check

PR #760, opened 2015-11-02
Build Refactor Step 1

cheeseng:
Moved project build code for regular and generated tests into RegularTests.scala and GenerateTests.scala.


cheeseng:
Will be submitted against different feature branch.

PR #773, opened 2015-11-12
Spec -> RefSpec Rename

cheeseng:
Added RefSpec and RefSpecLike, and deprecated Spec and SpecLike.


cheeseng:
Changes already in 3.0.x branch, no longer applicable.

PR #783, opened 2015-11-23
Replace <ul> element with <dl> in FunSuite scaladoc

gcberger:
A simpler workaround than previous ul-to-table one.


bvenners:
Didn't notice this one was a substitute for the previous one until after I had already merged in the previous one, so just did substitution by hand. I.e., this PR won, but the other got merged.

PR #805, opened 2015-12-16
Runnable Examples-js

cheeseng:
Make examples.js runnable by not copying examples that use java.io.File when generating examples.js.


cheeseng:
Closing because this is included in this newer PR: 

https://github.com/scalatest/scalatest/pull/1219

PR #807, opened 2015-12-17
Make the eventually construct understand failed assumptions

cb372:
When an `assume` fails, it throws a `TestCanceledException` to cancel the test. `eventually` should understand this and rethrow the exception instead of retrying.

See the new test for an example of what I'm talking about.


plinlor:
@cla-bot[bot] check

Issue #630, opened 2015-06-27
report absolute file and line number of failing test


fommil:
Many text editors can perform automatic "jump to error" and so on if the absolute file and line number is included in a warn/error message.

It would be incredibly useful for emacs/Vim/sublime/atom integration if failing scalatests were reported with full path and line number (emacs even supports column number).

The failing line in the test need not be automatically discovered (unrolling the stacktrace can be a pretty tough job), it is enough to have the line that begins the test.

For example, today if my test fails I see this:

```
[info] - should marshal DebugLocation *** FAILED ***
[info]   SexpCons(SexpSymbol(:return),SexpCons(SexpCons(SexpSymbol(:ok),SexpCons(SexpCons(SexpSymbol(:type),SexpCons(SexpSymbol(reference),SexpCons(SexpSymbol(:object-id),SexpCons(SexpString(57),SexpNil)))),SexpNil)),SexpCons(SexpNumber(666),SexpNil))) was not equal to SexpCons(SexpSymbol(:type),SexpCons(SexpSymbol(reference),SexpCons(SexpSymbol(:object-id),SexpCons(SexpString(57),SexpNil)))) (SwankFormatsSpec.scala:20)
[info]   org.scalatest.exceptions.TestFailedException:
[info]   at org.scalatest.MatchersHelper$.newTestFailedException(MatchersHelper.scala:160)
[info]   at org.scalatest.Matchers$AnyShouldWrapper.shouldBe(Matchers.scala:6408)
... big stack trace ...
```

Note the failing line 20 in the error message is just some utility, not the actual test, and the filename is not absolute. We could potentially infer the filename using ENSIME, but that means this won't work without an ENSIME server.

but if it was like this, then I'd be able to jump to that failing test easily:

```
[info] - should marshal DebugLocation *** FAILED *** /path/to/my/file/SwankFormatsSpec.scala:100
...
```

Then the filepath and line can be extracted with a pattern match.

Some people might find it useful to have the filename/line on ignored tests as well.

I've tried my best to get something working in https://github.com/hvesalai/sbt-mode/pull/29 but my changes only allow the user to jump the cursor to the failure message, it doesn't open up the failed test's source code.

bvenners:
Hi Sam,

Are you familiar with the Location API in ScalaTest's event model? Its purpose is to allow IDEs to hop to the relevant line of code for events for which that is useful. It is an ADT defined here:

[...]

PR #820, opened 2015-12-29
Machine readable location information for Emacs

fommil:
This is my best attempt at #630 but it's probably not a particularly clean implementation and I'm not really able to test it properly because my machine is not powerful enough to compile/test scalatest.

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/tools/StringReporter.scala
+++ b/scalatest/src/main/scala/org/scalatest/tools/StringReporter.scala
@@ -781,9 +781,19 @@ private[scalatest] object StringReporter {
         val tff: Vector[Fragment] = fragmentsOnError(Resources.failedNote, Resources.testFailed _, message, throwable, formatter, Some(suiteName), Some(testName), duration,
             presentUnformatted, presentAllDurations, presentShortStackTraces, presentFullStackTraces, AnsiRed)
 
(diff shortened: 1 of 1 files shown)


bvenners:
Awesome thanks. We'll run the tests. I also want to try out the whole ENSIME experience Is emacs the best way to try it, and is this how I would get going:

https://github.com/ensime/ensime-emacs/wiki/Quick-Start-Guide

I never had enough stick-to-it-ive-ness to learn emacs. Always switched back to vi before I crossed the threshold into day to day usability. Think how much more I could have accomplished if I'd only had the patience! But perhaps it is not too late. I will need to relearn at least a few keypresses to test this.

fommil:
yes that's probably the best guide, but I guess it assumes Emacs experience, so you may find yourself digressing into general Emacs tutorials (I won't discourage such honourable behaviour!). We're hoping to update the docs over the coming months to make the experience better.

fommil:
I don't understand how my changes could have broken the build, since they shouldn't be executed :-/

I'm surprised that travis can run your tests! Scalatest requires a lot more horsepower than ENSIME, and travis wasn't able to handle ENSIME so we moved to a self-hosted drone.

I should emphasise that I haven't been able to run this branch yet, so I can't confirm that it does what I originally wanted. But hopefully it makes sense?

PR #824, opened 2016-01-13
Immediate Note & Alert in SBT

cheeseng:
Make note and alert to print directly to standard output immediately without waiting for the suite to complete.


bvenners:
Closing this because we discovered:

logBuffered in Test := false

PR #854, opened 2016-03-14
WIP: Prettifier Refactoring [DO NOT MERGE]

cheeseng:
This refactoring is bigger than expected, so this PR is created earlier so that the review process can already.


cheeseng:
We'll use a better approach for SourceInfo and Prettifier.

Issue #855, opened 2016-03-15
Test compilation fails when attempting to use "assertDoesNotCompile"


pawelkrol:
Observed below described problem with the latest `ScalaTest` version `2.2.6`.

Since the following code does not compile...

```
scala> val i: Int = null
<console>:10: error: an expression of type Null is ineligible for implicit conversion
       val i: Int = null
                    ^
```

...I would have expected making an assertion regarding the same code defined in a spec file to pass:

```
import org.scalatest.FunSpec

class MySpec extends FunSpec {
  describe("class Null") {
    it("cannot assign a null value to an integer variable") {
      assertDoesNotCompile("val i: Int = null")
    }
  }
}
```

However, it does not. It does not even compile the test itself:

```
> test
[info] Compiling 1 Scala source to /home/djgruby/scala-playground/target/scala-2.11/test-classes...
[error] <macro>:1: an expression of type Null is ineligible for implicit conversion
[error] { val i: Int = null }
[error]                ^
[error] one error found
[error] (test:compileIncremental) Compilation failed
[error] Total time: 0 s, completed Mar 15, 2016 9:12:48 PM
```

Is it an issue with a package, or am I missing something in regard to how `assertDoesNotCompile` should work?

Many thanks!

bvenners:
Looks like a bug in assertDoesNotCompile, because assertDoesNotCompile should itself always compile.

pawelkrol:
Thanks a lot for your confirmation. I am looking forward to a new release version containing a bug fix.

cheeseng:
@pawelkrol @bvenners Just a quick update, fix for this is on its way: 

https://github.com/cheeseng/scalatest/tree/compile-macro-fix

Once all related problems are fixed I think we should backport the fixes to 2.2.x too.

pawelkrol:
Thank you for an update. Very much appreciated!

daniel-kullmann:
Shouldn't htis issue be closed, then?

PR #859, opened 2016-03-21
CompileMacro Fix

cheeseng:
Fixed CompileMacro problem when used with 'val i: Int = null'.


cheeseng:
This is the fix for the following issue: 

https://github.com/scalatest/scalatest/issues/855

PR #868, opened 2016-04-12
Prettifier & SourceInfo Refactoring

cheeseng:
Refactoring to enable support of customer Prettifier and use SourceInfo for stack depth resolving.

### DIFF
--- a/scalatest.js/src/main/scala/org/scalatest/exceptions/StackDepthExceptionHelper.scala
+++ b/scalatest.js/src/main/scala/org/scalatest/exceptions/StackDepthExceptionHelper.scala
@@ -43,8 +43,11 @@ private[scalatest] object StackDepthExceptionHelper {
     getStackDepth(sde.getStackTrace, fileName, methodName, adjustment)
   }
 
+  def isMatch(ele: StackTraceElement, pos: source.Position): Boolean =
+    Option(ele.getFileName).map(retrieveFileName) == Some(pos.fileName) && ele.getLineNumber == pos.lineNumber
+
   def getStackDepth(stackTrace: Array[StackTraceElement], pos: source.Position): Int = {
-    val idx = stackTrace.indexWhere (e => Option(e.getFileName).map(retrieveFileName) == Some(pos.fileName) && e.getLineNumber == pos.lineNumber)
+    val idx = stackTrace.indexWhere (e => isMatch(e, pos))
     if (idx >= 0)
       idx
     else {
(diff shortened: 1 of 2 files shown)


project/GenFactories.scala
```diff
@@ -533,7 +532,7 @@ $endif$
    *                 ^
    * </pre>
    */
-  def and(containWord: ContainWord): AndContainWord = new AndContainWord
+  def and(containWord: ContainWord)(implicit prettifier: Prettifier, pos: source.Position): AndContainWord = new AndContainWord(prettifier, pos)
```
bvenners: Why do and and or need the implicit prettifier and position here? Why can't the one captured by should be passed down?

common-test/src/main/scala/org/scalatest/SharedHelpers.scala
```diff
@@ -828,7 +829,7 @@ object SharedHelpers extends Assertions with LineNumberHelper {
         case map: GenMap[_, _] => element.asInstanceOf[Tuple2[_, _]]._1
         case genTrv: GenTraversable[_] => getIndex(xs, element)
       }
-    Array(indexOrKey.toString, decorateToStringValue(element))
+    Array(indexOrKey.toString, decorateToStringValue(Prettifier.default, element))
```
bvenners: Why is Prettifier.default used here? Why isn't a captured Prettifier passed down?

scalactic-macro/src/main/scala/org/scalactic/BooleanMacro.scala
```diff
@@ -802,13 +802,13 @@ private[org] class BooleanMacro[C <: Context](val context: C, helperName: String
    *   [code generated from callHelper]
    * }
    */
-  def genMacro[T](booleanExpr: Expr[Boolean], methodName: String, clueExpr: Expr[Any]): Expr[T] = {
+  def genMacro[T](booleanExpr: Expr[Boolean], methodName: String, clueExpr: Expr[Any], prettifierExpr: Expr[_], posExpr: Expr[_]): Expr[T] = {
```
bvenners: CS: Why Expr[_] for prettifierExpr and posExpr?

cheeseng: The reason is that while BooleanMacro sits in scalactic-macro, Prettifier and Position sits in scalactic, which depends on scalactic-macro.  scalactic-macro has no access to Prettifier and Position at that point.

project/GenFactories.scala
cheeseng: I'll check back this one.

cheeseng:
Majority changes are merged into 3.0.x already, the only one commit left is: 

https://github.com/scalatest/scalatest/pull/868/commits/f0c99b3a6448e26b4dbb4d406f4a4ca3665be6fe

We'll include this changes in a later PR.

PR #870, opened 2016-04-21
Don't call wait(0, 0) in AsyncAssertions.awaitImpl

zsxwing:
`wait(0, 0)` and `wait(0)` will wait forever until notified. `awaitImpl` should not call `wait` with 0 otherwise it will block forever and never timeout if `notifyAll` is not called.

This patch fixes `awaitImpl` by always calling `wait` with non-zero values.


plinlor:
@cla-bot[bot] check

PR #871, opened 2016-05-03
Fix EndWith regex failures

kevinshieh:
Was using `endWith` in scalatest and found the following bug:

`scala> "ab20" should endWith regex ("\d+")

scala> "a10b20" should endWith regex ("\d+")
org.scalatest.exceptions.TestFailedException: "a10b20" did not end with a substring that matched the regular expression \d+
  at org.scalatest.MatchersHelper$.newTestFailedException(MatchersHelper.scala:160)
  at org.scalatest.Matchers$ResultOfEndWithWordForString.regex(Matchers.scala:2324)
  at org.scalatest.Matchers$ResultOfEndWithWordForString.regex(Matchers.scala:2295)
  ... 38 elided`

The `endWith` implementation looks like this:
`val allMatches = rightRegex.findAllIn(left)
allMatches.hasNext && (allMatches.end == left.length)`

This fails because `findAllIn` doesn't find all, it returns an iterator so doing allMatches.end just grabs the first match. In the above example, this would grab the chunk that says "10" and only be halfway through the string. This then fails the check `allMatches.end == left.length` since the regex didn't have the chance to search until the end of the string.

My fix is as follows:
`val pMatcher = rightRegex.pattern.matcher(left)
while (!pMatcher.hitEnd) { pMatcher.find }
pMatcher.hitEnd && (pMatcher.end == left.length)`

I use a matcher and iterate through until the end. When we reach the end, THEN we check if the ending indices match up. Unfortunately, I don't think there is a way to reverse regex, it still has to search from beginning to end.

Apologies for all of the white space changes, I have "remove trailing whitespace" on Sublime turned on. Can revert the whitespace changes if desired.

### DIFF
--- a/scalatest-test/src/test/scala/org/scalatest/words/EndWithWordSpec.scala
+++ b/scalatest-test/src/test/scala/org/scalatest/words/EndWithWordSpec.scala
@@ -19,23 +19,23 @@ import org.scalatest._
 import Matchers._
 
 class EndWithWordSpec extends FreeSpec with FileMocks {
-  
+
   "EndWithWord " - {
-    
+
     "should have pretty toString" in {
       endWith.toString should be ("endWith")
     }
-    
+
     "apply(String) method returns Matcher" - {
-      
+
       val mt = endWith ("er")
-      
+
       "should have pretty toString" in {
         mt.toString should be ("endWith (\"er\")")
       }
-      
+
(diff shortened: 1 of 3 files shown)


cheeseng:
Resubmitted as: 

https://github.com/scalatest/scalatest/pull/1304

Issue #880, opened 2016-05-17
Extended Page Object Pattern Support (Geb like)


drieks:
Hello @bvenners,

we are using the page object pattern similar to the implementation provided by geb in our scalatest webpage tests and we would ask if you are interested in an contribution of this code to scalatest.

Here is an example:

```
package org.scalatest.examples.selenium

import org.scalatest.FunSpec
import org.scalatest.concurrent.Eventually
import org.scalatest.selenium.Chrome
import org.scalatest.selenium.DomainPage
import org.scalatest.selenium.Firefox
import org.scalatest.selenium.PageObject
import org.scalatest.selenium.WebBrowser
import org.scalatest.selenium.WebBrowserLauncher

/**
 * It is recommended to create a DomainPage if you are testing more then one page in the same domain.
 */
trait GoogleDomainPage extends DomainPage with PageObject with WebBrowser {
  /**
   * For this example the domain is hardcoded. But you can also do something like:
   * val domain = Option(System.getenv("DEV_SERVER_PREFIX")).getOrElse("https://www.google.com")
   */
  val domain = "https://www.google.com"
}

/**
 * This is the "Page Object Pattern" object for the google search page.
 *
 * A Page Object should not contain any state.
 */
case object GoogleSearchPage extends GoogleDomainPage {
  val path = "/"

  /**
   * a content object is required, it should contain functions returning elements from the page
   */
  object content {
    // in this example, we only need to access element with name "q"
    def q = textField("q")
  }

  override def at(): Unit = {
    assert(pageTitle == "Google")
  }
}

/**
 * WebBrowserLauncher will use @WrapWith to create Test instances as needed.
 *
 * bolierplate code like this is not needed:
 * <pre class="stHighlight">
 * class GoogleSearchSpecWithChrome extends GoogleSearchSpec with Chrome
 * class GoogleSearchSpecWithSafari extends GoogleSearchSpec with Safari
 * class GoogleSearchSpecWithInternetExplorer extends GoogleSearchSpec with InternetExplorer
 * class GoogleSearchSpecWithFirefox extends GoogleSearchSpec with Firefox
 * </pre>
 */
abstract class BaseSpec extends FunSpec with WebBrowserLauncher with Eventually {
  def Browsers = Seq(Firefox, Chrome)
}

class GoogleSearchPageObjectSpec extends BaseSpec {
  describe("google.com") {
    it("should change its title based on the term searched") {
      // Cancel test when cannot access google.com
[...]

PR #884, opened 2016-05-20
refs #880 AnnotationHelper to support @WrapWith on traits

drieks:
This PR will add a new AnnotationHelper what is used to find all @WrapWith Annotations, even on traits. We need this to add @WrapWith to a DriverLauncher trait that should start the actual test for different Browsers.


drieks:
Hello @bvenners ,

is there any way to get this PR into scalatest? We really need the possibility to add the @WrapWith annotation to traits.

If you have any change requests please let me know.

drieks:
there is another bug on @WrapWith handing, see #931

PR #887, opened 2016-05-23
Use the value property of `TagAnnotation` for the tag name.

gshakhn:
`TagAnnotation` has a handy `value` field, but it was ignored when
figuring out tag names. If `value` is set, use it for the tag name.
If it's not set, use the annotation class name instead.

This change is not backwards compatible, but the migration path
should be just changing which tag name you filter on.

* Create helper `AnnotatedSuiteToTagNames` and
  `AnnotationClassToTagName` objects to DRY up the annotation -> tag
  name logic.
* Use the logic in places where were were doing something like:

```
annotationClass = a.annotationType
if annotationClass.isAnnotationPresent(classOf[TagAnnotation])
```


plinlor:
@cla-bot[bot] check

PR #912, opened 2016-06-14
Upgraded to use SBT 0.13.11

VEINHORN:


cheeseng:
This is already in 3.1.x, we'll still use 0.13.9 for 3.0.x.

PR #924, opened 2016-07-04
Scala 2.12.0-M5 Failed Tests Fix

cheeseng:
Fixed failing tests when build using Scala 2.12.0-M5.


monkey-mas:
Hi @cheeseng, cc: @bvenners

AFIU, the problem is that we haven't imported `scala.collection.JavaConverters._` as the error message indicates.
https://travis-ci.org/scalatest/scalatest/jobs/142128350#L673

Because we generate these test files automatically[1], I've added the import to the base classes such as `ListShouldContainNoneOfLogicalAndSpec.scala`[2]. On my local machine, it seems the compilation succeeded and tests passed with this change. What do you think?

[1] https://github.com/scalatest/scalatest/blob/3.0.x/project/GenContain1.scala#L126
[2] https://github.com/monkey-mas/scalatest/commit/9e6292688a11a13179d985435297d298bc9d833d

cheeseng:
This PR is now out-dated, tests already passing in current 3.0.x using scala 2.12.4.

PR #938, opened 2016-07-16
[POST-3.0.0] SBT Framework -P support

cheeseng:
Added -P support to Framework, which allows specifying number of threads to use for thread pool used to run ParallelTestExecution suites.

Specifying a negative thread count will make it to use Executors.newCachedThreadPool.

This changes should be cherry-picked into 3.1.x and subsequently pulled into 3.2.x.


scalatest/src/main/scala/org/scalatest/tools/Framework.scala
```diff
@@ -660,7 +661,11 @@ class Framework extends SbtFramework {
         }
       }
 
-    val poolSize = Runtime.getRuntime.availableProcessors * 2
+    val poolSize =
```
drieks: can we check for -1 here and use ```Executors.newCachedThreadPool(threadPool)``` in this case?

cheeseng: @drieks sorry for a year of delay in responding, but yes I agree that's a good suggestion, I have added it in the latest commit in this PR.

bvenners:
Hi Chee Seng,

Please use C to indicate cached thread pool instead of a negative number. So these should work:

-P // Parallel execution
-P8 // Parallel execution with 8 threads in the thread pool
-PS8 // Parallel execution with 8 threads in the thread pool with suite sorting
-PC // Parallel execution with cached thread pool
-PSC  // Parallel execution with cached thread pool and suite sorting
-PCS // Parallel execution with cached thread pool and suite sorting

But a negative number or zero should give an error message:

-P-1
-P0
-PS-3
-PS0

Thanks.

Bill

PR #939, opened 2016-07-21
[POST-3.0.0] Differ for 'should equal' and 'shouldEqual'

cheeseng:
Added Differ support to should equal and shouldEqual matchers.


cheeseng:
Superseded by https://github.com/scalatest/scalatest/pull/951

PR #945, opened 2016-07-29
Unit Assertion Result Type for JUnitSuite and TestNGSuite

cheeseng:
Refactored Assertions to make JUnitSuite and TestNG's assert and assume method to return Unit instead of Assertion.


cheeseng:
Bill decided not to do this anymore.

Issue #517, opened 2015-02-20
Support for fluid element search with selenium


damluar:
Currently it's inconvinient to search elements relative to another element. For example, I have a table
`val newsHeadlines: XPathQuery = xpath("//*[@id ='gwt-debug-newsHeadlines']//table")`

To find `nth` row inside this table I have to do this:
find(newsHeadlines)
      .map(_.underlying)
      .map(_.findElement(By.xpath("//tr[" + position + "]")))

which seem really long and obscure to me. Is there a better way to do this?
I would be nice if you could chain queries one after another.

azzoti:
Agree 100%, the API seems flawed in this respect. 

find and findAll are methods on WebBrowser, find and findAll should also be available on WebBrowser.Element so you can say something like
```scala
find(newsHeadlines).map(_.find(xpath("//tr[" + position + "]"))) 
```
Its crazy to need to access Element.underlying - you might as well just use raw selenium-java .

Note that you have this in selenium-java:
```scala
public interface WebDriver extends SearchContext ...
public interface WebElement extends SearchContext ...
public interface SearchContext {  
    List<WebElement> findElements(By by);
    WebElement findElement(By by);
    ...
```
So in selenium-java you can just say 
```java
d.findElement(xpath("//*[@id ='gwt-Headlines']//table")).findElement(xpath("//tr[" + position + "]"))
```

IMO something similar should to be done in the Scala wrappers

azzoti:
(I'm not suggesting that the Option part of `def find(...) : Option[Element]` stuff is thrown away either, just that find and findAll exist on Element too)

jducoeur:
+1 -- I'm working in a fairly complex page, and it would be *much* easier (and produce much clearer test code) if I could think in terms of navigation within the Elements, rather than globally.

jducoeur:
It's also worth noting: I could implement this myself as an implicit class, except that `createTypedElement()` is private.  Is there a reason for that to be private?  I mean, yeah, purity, but unless the Scala DSL covers absolutely *everything* the Selenium does, it seems like it would be useful to make this function available as a workaround.

(Indeed, I'm finding that I can't even successfully copy the relevant code from WebBrowser into my own harness, because Element is sealed...)

NikolajsCmihuns:
[...]

PR #952, opened 2016-08-07
Support for fluid element search with selenium #517

andyglow:
This will allow scala users to do `find`/`findAll` within specified element only.
This feature was requested here - https://github.com/scalatest/scalatest/issues/517

### DIFF
--- a/scalatest-test/src/test/scala/org/scalatest/selenium/WebBrowserSpec.scala
+++ b/scalatest-test/src/test/scala/org/scalatest/selenium/WebBrowserSpec.scala
@@ -29,6 +29,7 @@ import org.openqa.selenium.safari.SafariDriver
 import org.scalatest.Args
 import org.scalatest.FunSpec
 import org.scalatest.Matchers
+import org.scalatest.OptionValues._
 import org.scalatest.ParallelTestExecution
 import org.scalatest.ScreenshotOnFailure
 import org.scalatest.SharedHelpers.SilentReporter
@@ -675,6 +676,27 @@ class WebBrowserSpec extends JettySpec with Matchers with SpanSugar with WebBrow
     }
   }
 
+  describe("find in element") {
+    def head = {
+      val head = find(cssSelector("head"))
+      head should be ('defined)
+      head.value
+    }
+    it("should return None if specified item not found") {
+      go to (host + "index.html")
+      head.find("something") should be(None)
+    }
+    it("should return a defined Option[Element] containing an instance of Element") {
+      go to (host + "index.html")
+      head.find(cssSelector("title")) match {
+        case Some(element) =>
+          element.text should be ("Test Title")
+        case other =>
+          fail("Expected Some(element), but got: " + other)
+      }
+    }
+  }
+
   describe("findAll") {
     it("should return an empty Iterator if specified item not found") {
       go to (host + "index.html")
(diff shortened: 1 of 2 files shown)


drieks:
Hi @andyglow,

please have a look at https://www.pageobject.org/. This is a extension for scalatest based on the code you have modified with many extensions. Support for single/multiple element find commands is also included. A overview presentation about the API is linked on the homepage. Feel free to contact me if you have any questions.

Dennis

p-kozlov:
@drieks by default pageobject library uses firefox for tests if I extend PageObjectSuite like here https://www.pageobject.org/downloads/PageObjectTour.pdf
Do you have any examples how to customize that? Should I create my own trait from the scratch with all which is needed, or is there something else to extend?

drieks:
Hi @p-kozlov,

> Looks like it should be enough to mixing PageBrowser trait and override webDriver with anything I want - correct?

This is not needed, you can run each test class in different browsers at the same time.

There are some DriverFactory classes that are responsible for creating the browser window.

There are two "collections" (called DriverFactoryList) of them:
* DefaultDriverFactoryList  is a collection of DriverFactory's that are opening the browser window directly on the desktop. (see https://github.com/agido/pageobject/blob/ed3f2a47fcd8a3a99f2c0e434e639003a5db87a6/core/src/main/scala/org/pageobject/core/driver/DefaultDriverFactoryList.scala#L75-L76)
* DefaultVncDriverFactoryList is a collection of DriverFactory's that are first launching a VNC Server and starting the browser inside of the server (recommended for linux developers and linux test servers) https://github.com/agido/pageobject/blob/ed3f2a47fcd8a3a99f2c0e434e639003a5db87a6/core/src/main/scala/org/pageobject/core/driver/vnc/DefaultVncDriverFactoryList.scala#L69-L70

As you can see there is already FirefoxDriverFactory and ChromeDriverFactory in the DefaultDriverFactoryList (and also others).

There are some checks which DriverFactoy should be used (e.g. InternetExplorerDriverFactory makes no sense when running on linux or osx). Another way is use environment variables to control how much browsers will be started (FIREFOX_LIMIT=0 will prevent starting any firefox browser, TEST_LIMIT=4 will allow up to 4 browsers of any type, CHROME_LIMIT=2 will allow up to two chrome browsers...)

if you want to use the DefaultVncDriverFactoryList  add the environment var RUN_WITH_DRIVERS=org.pageobject.core.driver.vnc.DefaultVncDriverFactoryList.

But back to your problem, if you want to use chrome and pageobject won't start chrome please try to add "-Dwebdriver.chrome.driver=/path/to/chromedriver" to your JVM arguments. (if you want to use only chrome and not firefox add also env FIREFOX_LIMIT=0) 

if you do not like this enviroment vars/JVM arguments you can annotate your test with @RunWithDrivers [org.pageobject.core.driver.RunWithDrivers] if you want to "hard code" the driver factory list. but when doing so you can't override the list using RUN_WITH_DRIVERS 

Example:
class MyDriverFactoryList extends DriverFactoryList(...)


Is chrome working when using -Dwebdriver.chrome.driver=... ?

drieks:
Your test should look like this:
https://github.com/agido/pageobject/blob/ed3f2a47fcd8a3a99f2c0e434e639003a5db87a6/examples/src/test/scala/org/pageobject/examples/tour04/Tour04.scala#L108

plinlor:
@cla-bot[bot] check

PR #962, opened 2016-08-26
Fix typecheck at ConfigMap

marcospereira:
This also split `getOptional` specs so they could run independently.

-----------------

I'm submitting this PR against 3.0.x branch because this is the version I'm currently using. Let me know if I'm doing something wrong here so I can promptly correct it.

### DIFF
--- a/scalatest-test/src/test/scala/org/scalatest/ConfigMapSpec.scala
+++ b/scalatest-test/src/test/scala/org/scalatest/ConfigMapSpec.scala
@@ -68,9 +68,102 @@ class ConfigMapSpec extends FunSpec {
       assert(caught.getMessage === Resources.configMapEntryNotFound("t"))
     }
 
-    it("should throw a TestCanceledException if a required entry has an unexpected type") {
+    describe("should provide a nice syntax to get") {
 
-      // Ensure supertype and subype is done correctly
+      it("Boolean values") {
+        // Ensure Boolean works
+        assert(cm.getRequired[Boolean]("boolean") === true)
+        assert(cm.getRequired[Boolean]("Boolean") === new java.lang.Boolean(true))
+        val caught2 =
+          intercept[TestCanceledException] {
+            cm.getRequired[Boolean]("string")
+          }
+
+        assert(caught2.getMessage === Resources.configMapEntryHadUnexpectedType("string", "class java.lang.String", "boolean", "aStringValue"))
+      }
+
+      it("Byte values") {
+        // Ensure Byte works
+        assert(cm.getRequired[Byte]("byte") === 1.toByte)
+        assert(cm.getRequired[Byte]("Byte") === new java.lang.Byte(1.toByte))
+        val caught3 =
+          intercept[TestCanceledException] {
+            cm.getRequired[Byte]("string")
+          }
+
+        assert(caught3.getMessage === Resources.configMapEntryHadUnexpectedType("string", "class java.lang.String", "byte", "aStringValue"))
+      }
+
+      it("Short values") {
+        // Ensure Short works
+        assert(cm.getRequired[Short]("short") === 1.toShort)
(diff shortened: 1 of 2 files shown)


marcospereira:
I've submitted #963 to fix some of the failing build jobs.

marcospereira:
Hey @cheeseng and @bvenners, 

Any feedbacks here? How can we move this forward?

cheeseng:
@marcospereira I am sorry for the delay in response we're currently tied up with some other tasks, we'll try to get back here to move forward as soon as possible.

marcospereira:
@cheeseng thanks for the heads up!

marcospereira:
Hey @cheeseng, any news about this PR (and also #963)?

cheeseng:
@marcospereira Sorry just finally able to study this PR, LGTM, we'll need to cherry pick this fix into 3.1.x as well.  Before we can merge in this PR, do you mind to print, sign and return the following CLA to  cla@artima.com: 

Thanks for the fix, we highly appreciate it!

marcospereira:
@cheeseng just sent the signed CLA.

Do you want me to submit another PR targeting branch 3.1 or will you guys manually port this change?

bvenners:
@marcospereira Can you rebase your lone commit and make it two commits? The fix is correct. The refactor of the tests, though, leaves them with incoherent English in the specification text (in the code). So I'd suggest we do that differently.

It is a tradeoff of FunSpec that although the words "describe" and "it" can help users, especially users new to BDD, express their test names in a way to sound like a specification, it imposes a structure on that text. Well written FunSpec's should follow that structure. For example, a top-level "describe" should be a subject (a noun). Nested "describe" clauses can then qualify that subject, such as describing specific situations the subject may find itself in. An "it" clause should be written whereby the "it" refers to the subject, and the next word is usually a verb (like should, can, or must), followed by the rest of the specification of behavior tested by that test. "It Boolean values" doesn't follow that. To structure it that way, you'd want to use a FreeSpec. In a FunSpec, you'd likely just put some duplication in the test names, like:

```
describe("A ConfigMap") {

  it("should provide a nice syntax to get Boolean values") {
    // ...
  }
  it("should provide a nice syntax to get Byte values") {
    // ...
  }
  // ...
}
```

Well, in addition to being overly verbose that's wrong in this case, because there is no separate syntax for Boolean or Byte values. By changing it to this what is lost is that that test code was supposed to test "A ConfigMap should throw a TestCanceledException if a required entry is missing".  

Lastly, I like to make a failing test before fixing a bug. I am not sure that happened here. The way this slipped under the radar is that the code threw a TestCanceledException when it wasn't supposed to, but that got buried in all the other canceled and pending tests that are happening. What we ought to do is state that it should not throw an exception when it shouldn't.  That would give us a failing test.

Also we should drop the phone number. I Googled and it is the number of my local water company which years ago must have called while I was coding and I wrote their number down in the code! So that one should definitely be removed.

Other than removing the phone number, I'd just leave the tests the same except adding failures because of the bug. There isn't a syntax for that in assertions as there is in matcher, so just do:

```
--- a/scalatest-test/src/test/scala/org/scalatest/ConfigMapSpec.scala
+++ b/scalatest-test/src/test/scala/org/scalatest/ConfigMapSpec.scala
@@ -72,7 +72,8 @@ class ConfigMapSpec extends FunSpec {
 
       // Ensure supertype and subype is done correctly
       assert(cm.getRequired[Apple]("apple") === apple)
-      assert(cm.getRequired[Fruit]("apple") === apple)
+      try { assert(cm.getRequired[Fruit]("apple") === apple) }
+      catch { case t: Throwable => fail(t) }
       val caught1 =
         intercept[TestCanceledException] {
           cm.getRequired[Apple]("fruit")
```

There are 3 places where that would be needed to turn the 3 canceled tests into failed ones.

PR #975, opened 2016-09-21
Deprecated either.{right,left}.value. Closes #972

### LINKED ISSUE #972: LeftProjection and RightProjection deprecated in 2.12.0-RC1
The following test:

``` scala
class T1 extends FunSuite with Matchers with EitherValues {
  test("EitherValues should work without deprecation warnings") {
      val r: Either[String, Int] = Right(5)
      val l: Either[String, Int] = Left("Teste")
      r.right.value shouldBe 5
      l.left.value shouldBe "Teste"
  }
}
```

With the following build.sbt:

``` sbt
name := "Test"

version := "1.0"

scalaVersion := "2.12.0-RC1"

libraryDependencies ++= Seq(
  "org.scalatest" %% "scalatest" % "3.0.0" % Test
)

scalacOptions ++= Seq(
  "-deprecation",                   //Emit warning and location for usages of deprecated APIs.
  "-Xfatal-warnings",               //Fail the compilation if there are any warnings.
  "-Xlint"                         //Enable recommended additional warnings.
)
```

Will fail:

```
Error:(7, 9) method right in class Either is deprecated (since 2.12.0): Either is now right-biased
      r.right.value shouldBe 5
Error:(8, 9) method left in class Either is deprecated (since 2.12.0): use swap instead
      l.left.value shouldBe "Teste"
```

Lasering:


SethTisue:
the deprecations will be reversed in 2.12.0-RC2 due to negative community feedback

they might still eventually go forward, so that doesn't necessarily mean this PR should be discarded — you could make a decision about what version of the code you like best, and perhaps leave feedback on https://github.com/scala/scala/pull/5135

wogan:
The deprecations are back as of 2.13 so can we get this patch applied please?

markus1189:
Because of the time that passed since the last comment a quick check.

The deprecation is still in place in `2.13.1`, does that mean it's going to stay?

SethTisue:
> The deprecation is still in place in 2.13.1, does that mean it's going to stay?

at present `.right` is deprecated (and I believe it will stay that way) but `.left` isn't, as per https://github.com/scala/scala/pull/8012

```
Welcome to Scala 2.13.1 (Java HotSpot(TM) 64-Bit Server VM, Java 1.8.0_202).
Type in expressions for evaluation. Or try :help.

scala 2.13.1> Right("foo").withLeft[Int]
res0: scala.util.Either[Int,String] = Right(foo)

scala 2.13.1> res0.left
res1: scala.util.Either.LeftProjection[Int,String] = LeftProjection(Right(foo))

scala 2.13.1> res0.right
                   ^
              warning: method right in class Either is deprecated (since 2.13.0): Either is now right-biased, use methods directly on Either
res2: scala.util.Either.RightProjection[Int,String] = RightProjection(Right(foo))
```

markus1189:
Okay thanks for the heads up :)

greg-wb-ap:
Would love to get this merged.
See also my comment on  #972 - I think it makes sense to keep `.left.value` and just add `.value` directly on Either.

clement-malaingre-ica:
Would really like to see this merged or some variation of it. Using the `toOption` workaround is just cluncky.

satorg:
I wonder, what exactly is missing in this PR that prevents it from being merged finally? It hangs open since 2016 already.

greg-wb-ap:
See also #1775 & #1746

plinlor:
@cla-bot[bot] check

rtyley:
This can be closed, as it was addressed in https://github.com/scalatest/scalatest/pull/1895.

PR #984, opened 2016-10-07
Build Aggregation Refactor (3.0.x)

cheeseng:
Refactored the build to aggregate test projects at scalatestApp and scalatestAppJS level to avoid dbuild cyclic dependency problem reported by Seth Tisue here: https://github.com/scala/scala-dev/issues/230#issuecomment-248774543 .

### DIFF
--- a/project/scalatest.scala
+++ b/project/scalatest.scala
@@ -334,7 +334,7 @@ object ScalatestBuild extends Build {
         "Bundle-DocURL" -> "http://www.scalactic.org/",
         "Bundle-Vendor" -> "Artima, Inc."
       )
-    ).dependsOn(scalacticMacro % "compile-internal, test-internal").aggregate(LocalProject("scalactic-test"))  // avoid dependency in pom on non-existent scalactic-macro artifact, per discussion in http://grokbase.com/t/gg/simple-build-tool/133shekp07/sbt-avoid-dependence-in-a-macro-based-project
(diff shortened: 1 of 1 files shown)


SethTisue:
thanks for merging this. this doesn't quite enable us to track your 3.0.x branch in the Scala 2.12 community build yet though, because the 3.0.x branch doesn't have a needed change to be compatible with 2.12.0-RC1 and the forthcoming 2.12.0-RC2, namely  https://github.com/adriaanm/scalatest/commit/49eeedde9beab0619dd8d6a0b46ef75f8e04c92a . can that change be merged onto 3.0.x? I think you might already have it on some other branch somewhere...?

bvenners:
@SethTisue The fix would be easy to merge into 3.0.x because it is just adding an explicit type annotation to an implicit val. Is that going to be a feature of 2.12 or do you plan to consider that a bug in RC1? If the latter then would you rather we *not* update 3.0.x so that it will start compiling once that is fixed in 2.12? Let us know what you prefer. We can merge in that fix into 3.0.x if you think that's best.

SethTisue:
We don't consider it a Scala bug, so I'm suggesting merging the change.

SethTisue:
@bvenners ping

bvenners:
@SethTisue is the change that's needed the one where we needed explicit result types on a couple vals?

SethTisue:
@bvenners yes

bvenners:
@SethTisue OK. I think it is this one: 

https://github.com/scalatest/scalatest/pull/990/commits/6ac5c15b560cb1ba8b82a65dfa1527ce294ddb0c

I'm merging in that PR now to both 3.0.x and 3.1.x.

bvenners:
@SethTisue I have merged that PR into both 3.0.x and 3.1.x. Let me know if you need anything else for the community build.

SethTisue:
I've got a test run queued up on Jenkins

PR #1002, opened 2016-11-08
GeneratorDrivenPropertyChecks using Generator

cheeseng:
GeneratorDrivenPropertyChecks now use our in-house built Generator, created ScalaCheckDrivenPropertyChecks to do what old GeneratorDrivenPropertyChecks was doing.


bvenners:
I merged this into feature-algebra-and-laws-2 instead of 3.1.x.

PR #1017, opened 2016-11-23
-N and -S Support

cheeseng:
-Added -N support to specify generator minSize default value.
-Added -S support to specify generator sizeRange default value.
-Avoid a race condition problem when multiple runner runs concurrently in SBT on the same JVM.


cheeseng:
Closing, changes in this PR has been resubmitted as: 

https://github.com/scalatest/scalatest/pull/1253
https://github.com/scalatest/scalatest/pull/1254
https://github.com/scalatest/scalatest/pull/1255
https://github.com/scalatest/scalatest/pull/1256

PR #1051, opened 2016-12-28
Create new class RegexString

gcberger:


scalactic-test/src/test/scala/org/scalactic/anyvals/RegexStringSpec.scala
```diff
@@ -0,0 +1,584 @@
...
+        }
+      }
+    }
+    it("should offer a codePointBefore method that is consistent with String") {
+      forAll { (regStr: RegexString, pint: PosInt) =>
+        whenever (regStr.length > 0) {
+          val idx = (pint % regStr.length) + 1
+
+          regStr.codePointBefore(idx) shouldEqual
+            regStr.value.codePointBefore(idx)
```
Telhar: i think it can't work as you want

line 132 - `shouldEqual` returns `Assertion`
line 133 - result of line 132 is ignored, and this line returns `Int`

bvenners: shouldEqual takes a parameter, so the compiler will look on the next line for it.

Telhar: yes, yes, my mistake.
my case is about `should be` construction, sorry.

bvenners:
I merged this in by cherry-picking the commit to 3.1.x and then pulling to feature-algebra-and-laws-2. Github doesn't think it is merged, so closing.

PR #1060, opened 2017-01-09
Cache WebElement's TagName and Type

valemin:
in order to reduce number of requests to browser when using createTypedElement


cheeseng:
Hi @valemin , 

We are sorry for the late review (almost a year), it seems that the changes are now implemented in WebBrowser via TagMeta already, in current 3.1.x branch.

We highly appreciate your help!

Thank you!

cc @bvenners

valemin:
Great, happy to hear that it's fixed.

PR #1068, opened 2017-01-20
Fix for Finite Float/Double when Merging

cheeseng:
Fixed failing FiniteFloatGeneratedSpec by adding correct NaN equality implementation.


cheeseng:
Already in feature branch.

PR #1069, opened 2017-01-20
Fix for Finite Float/Double when Merging

cheeseng:
Fixed failing FiniteFloatGeneratedSpec by adding correct NaN equality implementation.

Note: After pulling this into 3.1.x, the changes should be able to be pulled into feature branch and get the failing test passing.


cheeseng:
Branch off from wrong branch.

PR #1081, opened 2017-01-27
make HtmlReporter handle parallel test executions

LuboVarga:
I would like to improve reporter (now HtmlReporter) to be able to handle parallel test executions. These changes also does add git ignore for intelij idea directory.

PS: I know, that reporter will be now slower (it will wait till all test are finished and report generating will start after that time point), but it is able to handle parallel test executions. I hope I will make in near time replacement for html reporter, which will be fully parallel execution aware.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,5 @@ test-output/
 gentests/
 scalatest.js/target/
 *.swp
+.idea
+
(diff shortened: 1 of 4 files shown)


plinlor:
@cla-bot[bot] check

PR #1086, opened 2017-02-11
NumericString for Scala-js

cheeseng:
Skipped NumericString methods that are not supported on scala-js.


cheeseng:
Resubmitted for 3.1.x.

PR #1092, opened 2017-02-19
Asynchronous Inspectors

cheeseng:
Added asynchronous support to Inspectors.

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/Inspectors.scala
+++ b/scalatest/src/main/scala/org/scalatest/Inspectors.scala
@@ -263,7 +263,7 @@ trait Inspectors {
    * @param collecting the implicit <code>Collecting</code> that can transform <code>xs</code> into a <code>scala.collection.GenTraversable</code>
    *
    */
-  def forAll[ASSERTION](xs: String)(fun: Char => ASSERTION)(implicit collecting: Collecting[Char, String], asserting: InspectorAsserting[ASSERTION], prettifier: Prettifier, pos: source.Position): Unit = {
+  def forAll[ASSERTION](xs: String)(fun: Char => ASSERTION)(implicit collecting: Collecting[Char, String], asserting: InspectorAsserting[ASSERTION], prettifier: Prettifier, pos: source.Position): asserting.Result = {
     asserting.forAll(collecting.genTraversableFrom(xs), xs, false, prettifier, pos)(fun)
   }
 
--- a/scalatest/src/main/scala/org/scalatest/RecoverMethods.scala
+++ b/scalatest/src/main/scala/org/scalatest/RecoverMethods.scala
@@ -188,13 +188,18 @@ trait RecoverMethods {
   def recoverToSucceededIf[T <: AnyRef](future: Future[Any])(implicit classTag: ClassTag[T], exCtx: ExecutionContext, pos: source.Position): Future[Assertion] = {
     val clazz = classTag.runtimeClass
     future.failed.transform(
-      ex =>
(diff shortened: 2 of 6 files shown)


scalatest/src/main/scala/org/scalatest/Inspectors.scala
```diff
@@ -263,7 +263,7 @@ trait Inspectors {
    * @param collecting the implicit <code>Collecting</code> that can transform <code>xs</code> into a <code>scala.collection.GenTraversable</code>
    *
    */
-  def forAll[ASSERTION](xs: String)(fun: Char => ASSERTION)(implicit collecting: Collecting[Char, String], asserting: InspectorAsserting[ASSERTION], prettifier: Prettifier, pos: source.Position): Unit = {
+  def forAll[ASSERTION](xs: String)(fun: Char => ASSERTION)(implicit collecting: Collecting[Char, String], asserting: InspectorAsserting[ASSERTION], prettifier: Prettifier, pos: source.Position): asserting.Result = {
```
bvenners: Was this just a bug? Why was the result type Unit before?

scalatest/src/main/scala/org/scalatest/RecoverMethods.scala
```diff
@@ -188,13 +188,18 @@ trait RecoverMethods {
   def recoverToSucceededIf[T <: AnyRef](future: Future[Any])(implicit classTag: ClassTag[T], exCtx: ExecutionContext, pos: source.Position): Future[Assertion] = {
     val clazz = classTag.runtimeClass
     future.failed.transform(
-      ex =>
+      rawEx => {
+        val ex =
+          rawEx match {
+            case execEx: java.util.concurrent.ExecutionException => execEx.getCause
+            case other => other
+          }
```
bvenners: @cheeseng What was this change for? I can see you're grabbing the cause if it is an execution exception, but what was the situation in which this bug showed up?

scalatest/src/main/scala/org/scalatest/enablers/InspectorAsserting.scala
```diff
@@ -324,6 +326,264 @@ abstract class UnitInspectorAsserting {
     }
 }
 
+trait FutureInspectorAsserting {
```
bvenners: @cheeseng any reason this is a separate trait? The reason UnitInspectorAsserting is in a separate trait is to make it lowest priority implicit. If not, I'd prefer we eliminate this trait and just define this Impl class in the object and make it private to minimize API surface area.

scalatest/src/main/scala/org/scalatest/RecoverMethods.scala
cheeseng: Those exception that should be propogated thrown within the future are wrapped as ExecutionException, so here we extract them from ExecutionException and continue the checking.

Those failing tests are in the generated tests, to reproduce the problem, you can try remove the added code and run the following generated tests: 

rm -rf gentests
sbt genInspectorsTests/test

and see the result in the html reporter at target/html/index.html

scalatest/src/main/scala/org/scalatest/enablers/InspectorAsserting.scala
cheeseng: @bvenners Yes I think that makes sense, I'll try to make this adjustment next.

bvenners: Sorry I already did that and pushed it to 3.1.x.

PR #1102, opened 2017-02-27
InspectorAsserting Synchronization

cheeseng:
Cherry-picked InspectorAsserting's changes from feature-algebra-and-laws-2 branch.


cheeseng:
Will go with https://github.com/scalatest/scalatest/pull/1105 instead.

PR #1103, opened 2017-02-27
Asynchronous InspectorAsserting for feature-algebra-and-laws-2

cheeseng:
Asynchronous InspectorAsserting support for feature-algebra-and-laws-2.


cheeseng:
Will go with https://github.com/scalatest/scalatest/pull/1105 instead.

PR #1127, opened 2017-03-28
Wrap <script> sections of HtmlReporter in <![CDATA[]]>

biochimia:
One of the sections had an unescaped &, which trips up browsers when the
generated files are presented as XML/XHTML.

Preventively, this wraps the two longer <script> elements in a CDATA
block, even if only one currently needs a fix.


cheeseng:
Resubmitted for branch 3.0.x: https://github.com/scalatest/scalatest/pull/1307

PR #1139, opened 2017-04-17
path matchers

rvanheest:
ScalaTest has had file matchers (`file should exist`, `file shouldBe readable` and `file shouldBe writable`) for a long time. These are all based on the `java.io.File` class, which has been 'replaced' with the new `java.nio.files.Path` class as of Java 7. As I'm using this new-ish class, I found ScalaTest doesn't support this class yet. It's quite a simple addition, so I thought I'll submit this PR. If there is any functionality that I missed for `File` that also needs to be added for `Path`, let me know!

Added syntax should enable things like:

```scala
val path = Paths.get("path/to/some/file")

path should exist
path shouldNot exist
path should not (exist)

path shouldBe writable
path should be (writable)
path should not be writable
path shouldNot be (writable)

path shouldBe readable
path should be (readable)
path should not be readable
path shouldNot be (readable)

// as well as all equivalents with all(List(...))
```

### DIFF
--- a/scalatest-test/src/test/scala/org/scalatest/ShouldBeReadableSpec.scala
+++ b/scalatest-test/src/test/scala/org/scalatest/ShouldBeReadableSpec.scala
@@ -131,6 +131,73 @@ class ShouldBeReadableSpec extends FunSpec with Matchers {
 
     }
 
+    describe("when work with 'path should be (readable)'") {
+
+      it("should do nothing when path is readable") {
+        readableFile.toPath should be (readable)
+      }
+
+      it("should throw TestFailedException with correct stack depth when path is not readable") {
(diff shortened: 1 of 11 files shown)


rvanheest:
Hi, haven't heard anything on this PR. Is there something blocking? LMK!

cheeseng:
@rvanheest Sorry for the 'super' late reply, we're overwhelmed by works last year though we would want to catch up with PR asap.

One real blocker for this PR is really we still need to support Java 6 for Scala 2.10 and 2.11 build, imho this PR is a good enhancement if/once @bvenners decides to drop Java 6 support in future.  :)

plinlor:
@cla-bot[bot] check

PR #1154, opened 2017-05-04
Fixable for pendingUntilFixed

cheeseng:
Enhanced pendingUntilFixed with Fixable type class to make it behave correctly under asynchronous style traits.

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/fixture/AsyncWordSpecLike.scala
+++ b/scalatest/src/main/scala/org/scalatest/fixture/AsyncWordSpecLike.scala
@@ -50,7 +50,7 @@ StringVerbBlockRegistration, SubjectWithAfterWordRegistration}
  */
 //SCALATESTJS-ONLY @scala.scalajs.js.annotation.JSExportDescendentClasses(ignoreInvalidDescendants = true)
 @Finders(Array("org.scalatest.finders.WordSpecFinder"))
-trait AsyncWordSpecLike extends AsyncTestSuite with AsyncTestRegistration with ShouldVerb with MustVerb with CanVerb with Informing with Notifying with Alerting with Documenting { thisSuite =>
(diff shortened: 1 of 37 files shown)


scalatest/src/main/scala/org/scalatest/fixture/AsyncWordSpecLike.scala
```diff
@@ -50,7 +50,7 @@ StringVerbBlockRegistration, SubjectWithAfterWordRegistration}
  */
 //SCALATESTJS-ONLY @scala.scalajs.js.annotation.JSExportDescendentClasses(ignoreInvalidDescendants = true)
 @Finders(Array("org.scalatest.finders.WordSpecFinder"))
-trait AsyncWordSpecLike extends AsyncTestSuite with AsyncTestRegistration with ShouldVerb with MustVerb with CanVerb with Informing with Notifying with Alerting with Documenting { thisSuite =>
+trait AsyncWordSpecLike extends AsyncTestSuite with AsyncTestRegistration with ShouldVerb[Future[PendingStatement]] with MustVerb[Future[PendingStatement]] with CanVerb[Future[PendingStatement]] with Informing with Notifying with Alerting with Documenting { thisSuite =>
```
bvenners: I think we'll need to make an AsyncShouldVerb, AsyncMustVerb, and AsyncCanVerb where the signature of is has Future[PendingStatement], else this is a breaking change and couldn't go into a 3.x.y release. Plus it makes the types a bit simpler (at the cost of more classes).

bvenners: In addition we could potentially move pending and pendingUntilFixed out of Assertions (via a deprecation cycle). Put pending into trait/object Pending and pendingUntilFixed into trait/object PendingUntilFixed, and then just add an AsyncPendingUntilFixed that looks for Future[PendingStatement]. That would simplify the signatures of the two pendingUntilFixed methods.

plinlor:
@cla-bot[bot] check

Issue #1147, opened 2017-04-24
Race condition in the `Timed` trait implementation


axel22:
This has been biting me for a while now, so I decided to dig in:

```
[info] - reactor should ensure the ReactorStarted event *** FAILED ***
[info]   java.lang.IllegalStateException: Timer already cancelled.
[info]   at java.util.Timer.sched(Timer.java:397)
[info]   at java.util.Timer.schedule(Timer.java:193)
[info]   at org.scalatest.Timer.schedule(JavaClassesWrappers.scala:93)
[info]   at org.scalatest.enablers.Timed$$anon$3.timeoutAfter(Timed.scala:335)
[info]   at org.scalatest.enablers.Timed$$anon$3.timeoutAfter(Timed.scala:286)
[info]   at org.scalatest.concurrent.TimeLimits.failAfterImpl(TimeLimits.scala:239)
[info]   at org.scalatest.concurrent.TimeLimits.failAfterImpl$(TimeLimits.scala:233)
[info]   at io.reactors.concurrent.ReactorSystemTest.failAfterImpl(reactor-system-tests.scala:206)
[info]   at org.scalatest.concurrent.TimeLimits.failAfter(TimeLimits.scala:230)
[info]   at org.scalatest.concurrent.TimeLimits.failAfter$(TimeLimits.scala:229)
```

I believe that this is a race between the future created in line 306 here, which calls cancel in line 310:

https://github.com/scalatest/scalatest/blob/3.1.x/scalatest/src/main/scala/org/scalatest/enablers/Timed.scala#L310

and the `schedule` call on the same timer in line 335:

https://github.com/scalatest/scalatest/blob/3.1.x/scalatest/src/main/scala/org/scalatest/enablers/Timed.scala#L335

A similar race condition tends to happen with the line 309, where `task.cancel` is called.
Usually, this happens when the asynchronous test finishes very quickly, enabling it to *race* with the ScalaTest code.

The solution could be to add another promise in line 305, and use it to atomically decide on whether the task is canceled or scheduled:

```
val timerOwner = Promise[Unit]()
```

and then in line 309:

```
if (timerOwner.trySuccess(())) {
  task.cancel()
  timer.cancel()
}
```

and in line 335:

```
if (timerOwner.trySuccess(())) {
  timer.schedule(task, delay)
}
```

axel22:
@bvenners Do you agree with this analysis? Please let me know, and I will send a PR with a fix.

axel22:
[...]

PR #1155, opened 2017-05-06
Timed.scala Race Condition Fix

cheeseng:
Avoid race condition in Timed.scala as described in https://github.com/scalatest/scalatest/issues/1147 .


axel22:
LGTM. Not sure why [that test failed](https://github.com/scalatest/scalatest/blob/3.1.x/scalatest-test/src/test/scala/org/scalatest/concurrent/AsyncTimeLimitedTestsSpec2.scala#L45) in the Travis build, but it's possible that the [timeout should be increased](https://github.com/scalatest/scalatest/blob/3.1.x/scalatest-test/src/test/scala/org/scalatest/concurrent/AsyncTimeLimitedTestsSpec2.scala#L45).

PR #1158, opened 2017-05-09
Migration Manager (MiMa)

cheeseng:
Incorporated migration manager (MiMa) into build.


bvenners:
Next step for this enhancement:

https://github.com/scalatest/scalatest/issues/1266

PR #1160, opened 2017-05-11
EitherValues: Print value of Either when `right.value` or `left.value` fails

jjst:
This provides a more detailed error message when calling `.value` on an Either fails.

Currently, calling `.value` on a LeftProjection or RightProjection just fails with no details when the either is a right or a left respectively. This makes `EitherValues` clunky to use as it doesn't give enough details when a failure occurs. This PR changes the failure message so that it prints the actual value of the `Either`.

### DIFF
--- a/scalatest-test/src/test/scala/org/scalatest/EitherValuesSpec.scala
+++ b/scalatest-test/src/test/scala/org/scalatest/EitherValuesSpec.scala
@@ -38,7 +38,7 @@ class EitherValuesSpec extends FunSpec {
         }
       caught.failedCodeLineNumber.value should equal (thisLineNumber - 2)
       caught.failedCodeFileName.value should be ("EitherValuesSpec.scala")
-      caught.message.value should be (Resources.eitherLeftValueNotDefined)
+      caught.message.value should be (Resources.eitherLeftValueNotDefined(e))
     }
     
     it("should return the right value inside an either if right.value is defined") {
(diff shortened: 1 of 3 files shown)


lemastero:
I have seen (and done myself) lots of crazy workarounds, because testing Either does not have good error reporting. 

@bvenners @cheeseng can you use your superpowers - to make it happen :)

plinlor:
@cla-bot[bot] check

jjst:
Thanks for the ping; however, it looks like this patch has been superceded by #1738, opened 3 years later, which essentially does the same, albeit more comprehensively: it also provides better error messages for `Try`!

It looks like this fix made it into the [3.1.1](https://github.com/scalatest/scalatest/releases/tag/release-3.1.1) release.

As a result, I think this PR can be closed now.

Unfortunately though, I just tested and it seems this feature has not been ported to Scalatest 3.2.x. 

I am happy to submit a separate patch to port #1738 over to the latest version, provided a maintainer signals interest first, since I am a bit frustrated with this originally sitting idle for 3+ years - to be clear this is not a criticism of the maintainers who do a thankless, voluntary job for the community for free and have limited free time/energy to deal with the massive influx of PRs; My intention is simply to communicate I am happy to contribute back but first want to make sure that the PR will get attention before I commit to do the work.

cheeseng:
@jjst My sincere apology that we have missed this one, this one somehow just didn't get in.  I have cherry-picked changes in #1738  into my latest PR here: 

https://github.com/scalatest/scalatest/pull/1878

which is planned for 3.2.2, not long from now.  I have released 3.2.2-M1 but before the cherry-picks, and since https://github.com/scalatest/scalatest/issues/1876 reported that the M1 fixed the critical bug I think 3.2.2 that includes the cherry picks should be far away.

Again, I am sorry for missing one since 3 years ago.

PR #1168, opened 2017-06-17
JSON Reporting Support

cheeseng:
Added JSON support to events and StringReporter.


scalatest/src/main/scala/org/scalatest/tools/ScalaTestFramework.scala
```diff
@@ -328,6 +333,12 @@ class ScalaTestFramework extends SbtFramework {
       }
     }
 
+    protected def printNoColor(text: String): Unit = {
```
bvenners: Hi Chee Seng,

In Framework you call printNoColor (on line 610 of Framework.scal), but in ScalaTestFramework you do not. I think you forgot to call it, which means -oJ wouldn't work in versions of sbt prior to 0.13. But maybe it is time to drop support for that anyway. Regardless, we should add that here. I'll go ahead and merge this in, but please add that as a separate PR.

Bill

PR #1171, opened 2017-06-21
Initial version of ultimately, which is an async eventually.

bvenners:


scalatest/src/main/resources/org/scalatest/ScalaTestBundle.properties
```diff
@@ -582,6 +582,8 @@ insidePartialFunctionAppendNone=An exception was thrown,\n{0}inside {1}
 
 didNotEventuallySucceed=The code passed to eventually never returned normally. Attempted {0} times over {1}.
 didNotEventuallySucceedBecause=The code passed to eventually never returned normally. Attempted {0} times over {1}. Last failure message: {2}.
+didNotUltimatelySucceed=The fugure-producing code passed to ultimately never yielded a successful future. Attempted {0} times over {1}.
```
fsvehla: @bvenners fugure → future

PR #1179, opened 2017-07-06
Dependencies Version Update

cheeseng:
Bumped up version number for scala xml, combinator parser and scala-js.


cheeseng:
Superseded by https://github.com/scalatest/scalatest/pull/1214

PR #1182, opened 2017-07-11
scala.js 1.0.0-M1 support

### LINKED ISSUE #1178: Support scala.js 1.0.0-M1
it would be good to add support for 1.0.0 milestone. I cannot publish my own open source due to my dependency on scaltest

cquiroz:
This PR lets build scalatest with either scala.js 0.6.18 or with scala.js 1.0.0-M1. I'd need this to build my own projects that I wont to run with 1.0.0-M1. I published this locally and was able to build and run my own projects this way

There are a few things pending which I'm not sure how to do as the build is fairly complex:

* Cross build on 0.6.18 and 1.0.0-M1 on travis
* Cross publish for both versions

This fixes #1178

### DIFF
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -2,4 +2,4 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-pgp" % "0.8.3")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.7.0")
 
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.16")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.18")
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -2,4 +2,4 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-pgp" % "0.8.3")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.7.0")
 
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "0.6.18")
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.0.0-M1")
--- a/project/plugins.sbt
+++ b/project/plugins.sbt
@@ -2,4 +2,7 @@ addSbtPlugin("com.typesafe.sbt" % "sbt-pgp" % "0.8.3")
 
 addSbtPlugin("com.typesafe.sbt" % "sbt-osgi" % "0.7.0")
 
-addSbtPlugin("org.scala-js" % "sbt-scalajs" % "1.0.0-M1")
+val scalaJSVersion =
+  Option(System.getenv("SCALAJS_VERSION")).getOrElse("0.6.18")
+
+addSbtPlugin("org.scala-js" % "sbt-scalajs" % scalaJSVersion)
--- a/project/scalatest.scala
+++ b/project/scalatest.scala
@@ -23,7 +23,7 @@ object ScalatestBuild extends Build {
 
   val releaseVersion = "3.0.3"
 
-  val scalacheckVersion = "1.13.4"
+  val scalacheckVersion = "1.13.5"
 
   val githubTag = "release-3.0.3" // for scaladoc source urls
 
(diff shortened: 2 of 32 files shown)


cheeseng:
@cquiroz Thanks for taking this up!  I shall try this out ASAP, at the mean time, do you mind to print, sign and email back the following CLA to cla@artima.com :

Thanks for the effort, we highly appreciate it!

cquiroz:
i sent the CLA, thanks

cheeseng:
@cquiroz @bvenners I tried it out, unfortunately I couldn't get the tests running yet, for a yet to be known reason, it has link error when I tried to run tests: 

[error] Referring to non-existent class org.scalacheck.Arbitrary$
[error]   called from org.scalactic.anyvals.PosDoubleSpec.<init>()
[error]   called from org.scalactic.anyvals.PosDoubleSpec.<clinit>()
[error]   called from core module analyzer
[error] Referring to non-existent class org.scalacheck.Gen$Choose$
[error]   called from org.scalactic.anyvals.PosDoubleSpec.<init>()
[error]   called from org.scalactic.anyvals.PosDoubleSpec.<clinit>()
[error]   called from core module analyzer
[error] Referring to non-existent class org.scalacheck.Gen$
[error]   called from org.scalactic.anyvals.PosDoubleSpec.<init>()
[error]   called from org.scalactic.anyvals.PosDoubleSpec.<clinit>()
[error]   called from core module analyzer
[error] Referring to non-existent class org.scalacheck.Shrink$
[error]   called from org.scalactic.anyvals.PosDoubleSpec$$anonfun$1$$anonfun$apply$mcV$sp$4$$anonfun$apply$mcV$sp$22.apply()org.scalatest.compatible.Assertion
[error]   called from org.scalactic.anyvals.PosDoubleSpec$$anonfun$1$$anonfun$apply$mcV$sp$4$$anonfun$apply$mcV$sp$22.apply()java.lang.Object
[error]   called from org.scalatest.SuperEngine.registerFlatBranch(java.lang.String,scala.Function0,java.lang.String,java.lang.String,scala.Int,scala.Int,scala.Option)scala.Unit
[error]   called from org.scalatest.FlatSpecLike$$anon$2.apply(java.lang.String,java.lang.String,java.lang.String,org.scalactic.source.Position)org.scalatest.words.ResultOfStringPassedToVerb
[error]   called from org.scalatest.words.ShouldVerb$StringShouldWrapperForVerb$class.should(org.scalatest.words.ShouldVerb$StringShouldWrapperForVerb,java.lang.String,org.scalatest.words.StringVerbStringInvocation)org.scalatest.words.ResultOfStringPassedToVerb

I'll try to investigate further tomorrow.

FYI, to run tests for scala-js, you'll do: 

sbt scalatestAppJS/test

cquiroz:
It seems there are troubles with scalacheck. They are still at an older scala.js version but I'll try to move scalacheck forward too

cquiroz:
It seems there is also some dependency that uses scala.js 0.6.13. That explains at least one linking error on `commonTestJS`. I'm still trying to isolate it but the build is fairly complex

BTW any reason not to update the scala versions at the latest? It seems there are parts mixed between 2.11.8 and 2.11.11 as shown when calling `evicted`

cquiroz:
A little progress here. After digging a bit I found that there were certain dependency on scala.js 0.6.13 and 0.6.18 which caused one linking error. It was solved by removing the `module` setting on sbt.

Even after searching quite a bit I still don't know what it does or how it adds this shadow dependency. any ideas?

With this change `commonTestJS/fastOptJS` works but there is still an error related to scalacheck

cheeseng:
@cquiroz sorry for replying late, yes we should update 2.11.8 to 2.11.11, should be the right move for that.

I notice you submitted a PR to scalacheck, but not merged in yet, it means scalacheck did not release yet for scala-js 1.0.0-M1?  It sound it should fail with missing dependency instead, but weird that it didn't.  I wonder if there's a fallback to 0.6.18 during compile time but failed at linking time, not sure. @sjrd any idea what could be causing the behavior?

Thanks.

sjrd:
`%%%` would definitely not fallback on 0.6.x in a 1.0.0-M1 build. That would be totally useless since 1.x cannot read 0.6.x binaries. Maybe there's an explicitly written `_sjs0.6` somewhere?

cquiroz:
@cheeseng I think this PR will be a work in progress until `scalacheck` releases a version on `1.0.0-M1`. This works for me so far because I don't use `scalacheck` in my projects and thus linking doesn't fail

PR #1236, opened 2017-10-27
Smart Generators

cheeseng:
- Adjusted commands to run generated tests in README.md.
- Make all generators to generate files only when necessary.


cheeseng:
Failed travis build checked, failed due to flicker.

PR #1258, opened 2017-11-17
Smart GenAnyVals

cheeseng:
Made GenAnyVals.scala smarter to generate files only when necessary.


bvenners:
This is also pulled into 3.2.x.

PR #1280, opened 2017-11-29
WiP Towards support for Scala.js 1.x

sjrd:
Let's see whether the CI is as unhappy as my local setup, or whether I did not understand how to run this locally.

In any case, the contents of the PR can be useful for anyone who would like to pick up the support for Scala.js 1.x.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,3 +3,4 @@ test-output/
 gentests/
 scalatest.js/target/
 *.swp
+.idea
--- a/.travis.yml
+++ b/.travis.yml
@@ -14,7 +14,7 @@ jdk:
   - oraclejdk8
 
 dist: trusty
-sudo: false
+sudo: true
 
 sbt_args: -J-Xss10m
 
--- a/README.md
+++ b/README.md
@@ -117,6 +117,30 @@ If you would like to export a particular private key into a separate GPG file, y
 
 With Sonatype credentials and GPG file in place, you can now publish to Sonatype.
 
+Before publishing any patch release, binary compatibility with previous version should be checked, using Java 6 (for Scala 2.10 and 2.11):
+
+    $ sbt ++2.11.11 scalactic/package scalactic/mimaReportBinaryIssues
+    $ sbt ++2.11.11 scalatest/package scalatest/mimaReportBinaryIssues
+    $ sbt ++2.11.11 scalacticJS/package scalacticJS/mimaReportBinaryIssues
+    $ sbt ++2.11.11 scalatestJS/package scalatestJS/mimaReportBinaryIssues
+
+    $ sbt ++2.10.6 scalactic/package scalactic/mimaReportBinaryIssues
+    $ sbt ++2.10.6 scalatest/package scalatest/mimaReportBinaryIssues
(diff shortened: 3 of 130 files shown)


sjrd:
@cheeseng The only failure seems spurious.

I don't know if the current CI is sufficient to validate much of the Scala.js support.

In any case, the changes in this PR will be necessary before upgrading to Scala.js 1.x. The other change that is required is dropping Rhino (and use Node.js, the default). I have tried that but it seems that some stack-trace-related tests fail in that case. Maybe they're a bit too sensitive?

cheeseng:
@sjrd Currently our CI does not build for scala-js, except the examples for scala-js.

It has been a while back, if my memory serves me right, I tried to run our tests with node.js but it always seems to crash with not enough memory problem, I am curious if your branch is running currently with any settings required for node.js?

Thanks!

cheeseng:
@sjrd One reason scala-js build was not in our CI is that it requires much more memory, e.g. 6G and longer time to build and run (on rhino, which takes more than 50 mins), so in fact we left it out of CI currently simply due to we could work out a way to make it friendly to travis yet.  :)

sjrd:
No you're right. I don't really manage to run `scalatestTestJS/test` with Node.js, which consistently crashes with JS out of memory even with
```scala
      //jsDependencies += RuntimeDOM % "test",
      // Give more memory to Node.js, and deactivate source maps
      jsEnv := {
        import org.scalajs.jsenv.nodejs.NodeJSEnv
        new NodeJSEnv(
            NodeJSEnv.Config()
              .withArgs(List("--max_new_space_size=3000", "--max_old_space_size=3000")))
      },
```
`scalacticTestJS/test` manages to run a bunch of tests, then seems to hang. It uses one of my CPUs 100% but I don't know if it's making any progress.

cheeseng:
@sjrd Got a chance to try running your branch, with --max_old_space_size=3000 it freezed my desktop with 16G memory, I wonder if setting parallelTestExecution := false will help, I shall give it a shot tomorrow morning.

cheeseng:
@sjrd just sharing, scalacticTestJS/test took really long to run on rhino, it will eventually finished but really long, I am trying to recall how long it took in my last run, probably more than an hour I think.

cheeseng:
@sjrd fyi I just submitted a PR to your branch here: 

https://github.com/sjrd/scalatest/pull/1

cheeseng:
@bvenners @sjrd Since this branch now works in node-js env, which runs much faster than rhino, I suggest to merge this branch into 3.1.x at this point.

sjrd:
FWIW, I agree.

After that, I think the only thing blocking the migration to Scala.js 1.0.0-M2 is support from ScalaCheck. See https://github.com/rickynils/scalacheck/pull/375

bvenners:
This is also merged into 3.2.x.

sjrd:
@cheeseng ScalaCheck 1.14.0 has been released with support for Scala.js 1.0.0-M3, which unblocks forward progress on this, i.e., you should now be able to publish a ScalaTest for Scala.js 1.0.0-M3 as well.

PR #1281, opened 2017-12-04
Scala-js Link Problem Fix

cheeseng:
Changed to use parser combinator for parsing JSON in test, fixing link problem when running scalatest tests in scala-js.  

However, when using parser-combinator as a dependency, we still encountered problem in scala 2.10 and scala 2.13.  parser-combinator 1.0.6 (the latest released version) supports build  2.13 but the jar is empty for scala-js artifact, 1.0.7 is supposed to fix it but it is not released: 

https://github.com/scala/scala-parser-combinators/issues/130

And in scala 2.10 the case is worse, parser-combinator does not support scala 2.10 because it is included directly in scala distribution, but scala-js does not seem to implement it as when I tried to build it, it built but encountered link error when I tried to run it.

The solution in this PR is to include the source files of parser combinators directly under common-test, it works for both scala 2.10 and 2.13.

### DIFF
--- a/common-test.js/src/main/scala/scala/util/parsing/LICENSE.md
+++ b/common-test.js/src/main/scala/scala/util/parsing/LICENSE.md
@@ -1,0 +1,28 @@
+Copyright (c) 2002-2013 EPFL
+Copyright (c) 2011-2013 Typesafe, Inc.
+
+All rights reserved.
+
+Redistribution and use in source and binary forms, with or without modification,
+are permitted provided that the following conditions are met:
+
+    * Redistributions of source code must retain the above copyright notice,
+      this list of conditions and the following disclaimer.
+    * Redistributions in binary form must reproduce the above copyright notice,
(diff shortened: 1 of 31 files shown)


bvenners:
This is also pulled into 3.2.x.

gourlaysama:
FYI: scala-parser-combinators 1.0.7 has been [released][1] with fixed scala-js support for 2.13.0-M2.
 
And 1.1.0 is coming soon with 2.13.0-M3 support :-)

I'm a bit surprised that the scala-js version of scala-library_2.10 doesn't include parser-combinators though...

[1]: https://github.com/scala/scala-parser-combinators/releases/tag/v1.0.7

ashawley:
But Scala 2.13 eventually dropped `PagedSeq` in scala/scala#6198, so this will break in 2.13.0-M3.

Will this require in turn copying the sources for `PagedSeq` as well?

PR #1282, opened 2017-12-04
Fix  indicateFailure in the UnitPropCheckerAsserting

vladimir-popov:
I can't see reason why property-based test failed, because argument `optionalCause` always is None.


plinlor:
@cla-bot[bot] check

PR #1283, opened 2017-12-04
DifferSpec Fix for Scala 2.13

cheeseng:
Fixed a failing test in DifferSpec when builds with scala 2.13.


bvenners:
This has also been pulled into 3.2.x.

PR #1288, opened 2017-12-09
Improve documentation of exceptional behavior of futurevalue

### LINKED ISSUE #1172: scalatest.concurrent.ScalaFutures.futureValue throws TestFailedException for failed futures
The following test fails with Scalatest 3.0.1
```
        trait Foo {
          def bar: Future[Boolean]
        }
        
        val mockFoo = mock[Foo]
        when(mockFoo.bar).thenReturn(Future.failed(new IOException("baz")))

        an[IOException] shouldBe thrownBy {
          mockFoo.bar.futureValue
        }
```

> Expected exception java.io.IOException to be thrown, but org.scalatest.exceptions.TestFailedException was thrown

However, according to the documentation, `futureValue` should throw the failed future exception.

>      * @throws Throwable if once ready, the <code>value</code> of this future is defined as a
     *       <code>Left</code> (in this case, this method throws that same exception)

(https://github.com/scalatest/scalatest/blob/d40d278f2bb8e73431b942d87881fb6dbab75cb9/scalatest/src/main/scala/org/scalatest/concurrent/Futures.scala#L347)

alenkacz:
Current documentation is very misleading - multiple people are in an illusion, that using thrownBy with futures is allowed while at the same time, recoverToSucceededIf seems to be the correct approach.

This PR tries to fix the documentation to make this more obvious.

Fixes #1172

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/concurrent/Futures.scala
+++ b/scalatest/src/main/scala/org/scalatest/concurrent/Futures.scala
@@ -344,10 +344,9 @@ trait Futures extends PatienceConfiguration {
      * @param timeout the <code>Timeout</code> configuration parameter
      * @param interval the <code>Interval</code> configuration parameter
      * @return the result of the future once it is ready, if <code>value</code> is defined as a <code>Right</code>
-     * @throws Throwable if once ready, the <code>value</code> of this future is defined as a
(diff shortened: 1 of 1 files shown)


plinlor:
@cla-bot[bot] check

PR #1290, opened 2017-12-13
Don't discard underlying exception

amir:


amir:
The only failure doesn't seem to be related to the changeset.

evis:
@cheeseng could we merge it? I'm interested in this bugfix too.

plinlor:
@cla-bot[bot] check

PR #1291, opened 2017-12-16
Scala Native for 3.2

cheeseng:
Merging Scala Native Support in 3.1.x into 3.2.x


olafurpg:
Any estimate when this will be available for use? This unblocks our last dependency to support native in scalameta (https://github.com/scalameta/scalameta).

PR #1303, opened 2018-01-01
Fix build compilation when using coursier

2m:
I am using coursier in my global sbt configuration. The current scalatest master build does not compile without these changes. I am not completely sure what is causing the issue, but it seems that `io` package gets shadowed by `"io.get-coursier" % "sbt-coursier" % "1.0.0"` sbt plugin.

### DIFF
--- a/project/GenExamplesJS.scala
+++ b/project/GenExamplesJS.scala
@@ -16,7 +16,7 @@
 
 import sbt.IO
 
-import io.Source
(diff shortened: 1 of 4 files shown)


plinlor:
@cla-bot[bot] check

2m:
I think this PR is not relevant anymore. :)

PR #1313, opened 2018-01-22
Fix typo EitherValues

AlvaroCaste:
_Problemas_ is masculine, therefore it must be _Muchos_ instead of _Muchas_


plinlor:
@cla-bot[bot] check

PR #1331, opened 2018-03-01
don't use octal escapes

SethTisue:
they are deprecated and will be removed in Scala 2.13

this came up at https://github.com/scala/scala/pull/6324


cheeseng:
LGTM, we should pull this change into 3.1.x and 3.2.x branch after it is merged into 3.0.x.

SethTisue:
the Travis timeout appears unrelated

PR #1334, opened 2018-03-12
FeatureSpec Module

cheeseng:
Refactored FeatureSpec into its own package, pathing way to modularization in the future.


cheeseng:
@bvenners Fixed the conflicts against current 3.1.x, it is ready for your review again.

PR #1337, opened 2018-03-21
FlatSpec Module

cheeseng:
Refactored FlatSpec into its own package, pathing way to modularization in the future.


cheeseng:
Superseded by: 

https://github.com/scalatest/scalatest/pull/1551

PR #1338, opened 2018-03-22
Save failure cause in case of TryValues.success invocation on failed Try

evis:
Let's say, we invoke `someFun()` in test, which returns `Try[T]`,
and we expect, that `someFun()` will return `Success(_)`.

We write such code:

```
import org.scalatest.TryValues._
val x = someFun().success.value
```

But `someFun()` unexpectedly returns `Failure(e)`,
and test fails with something like that:

```
org.scalatest.exceptions.TestFailedException: The Try on which success was invoked was not a Success.
[stack trace...]
```

And there is no information about `e` from returned `Failure(e)`!

We need to manually write `throw someFun().failure.exception` and
rerun test to get reason of failure... That's so annoying.

This commit adds cause of failure to `TestFailedException` to avoid
this clumsy debugging.

### DIFF
--- a/scalatest-test/src/test/scala/org/scalatest/TryValuesSpec.scala
+++ b/scalatest-test/src/test/scala/org/scalatest/TryValuesSpec.scala
@@ -60,6 +60,7 @@ class TryValuesSpec extends FunSpec {
       caught.failedCodeLineNumber.value should equal (thisLineNumber - 2)
       caught.failedCodeFileName.value should be ("TryValuesSpec.scala")
       caught.message.value should be (Resources.tryNotASuccess)
+      caught.cause.value shouldBe an[Exception]
     }
   } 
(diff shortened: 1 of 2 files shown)


evis:
@bvenners could you review this PR, please?

evis:
@cheeseng @bvenners review PR please?

evis:
@bvenners just met you at Scala Days and recalled about this 1 year old PR :) could we merge it please? :)

evis:
@cheeseng @bvenners ping

evis:
@cheeseng @bvenners review please?

evis:
@cheeseng @bvenners review please? It's tiny three lines PR :(

evis:
@cheeseng @bvenners review please...

PR #1340, opened 2018-03-23
Fix build for Java 9

ashawley:
Java 10 is following closely behind.

- Updates sbt to 0.13.17
- Drop `CMSIncrementMode` from JVMOPTS
- ~~Use Scala 2.13.0-M3~~
- Add JDK9 to Travis

"Java 9 is dead, long live Java 9!"

PR #1356, opened 2018-04-24
Add source for PagedSeq from scala

ashawley:
It was deprecated in 2.12 and removed in Scala 2.13.0-M3, see scala/scala#6198.

This change was needed while working on java-9 support in #1340, but  I stopped short of updating to Scala 2.13.0-M3 for other reasons.   This is a change you'll need to get there.

I saw the solution in the past was to add the code to the scalatest repo (scala's JSON and the parser combinators, I recall).


ashawley:
Closing in favor of #1379.

Issue #1359, opened 2018-04-25
ShouldBe issue with Int type resolves incorrectly


dhinojosa:
Given:
```scala
val x:Any = 40
x shouldBe an [Int]
```

Result:
```
40 was not an instance of int, but an instance of java.lang.Integer
```

*Tech Stack*
-------------
JDK 10
Scala 2.12.4
ScalaTest 3.0.5

bvenners:
Merged PR that fixes this.

PR #1364, opened 2018-05-05
TypeMatcherHelper Enhancement

cheeseng:
Enhanced TypeMatcherHelper to handle primitive types checking.

https://github.com/scalatest/scalatest/issues/1359

Note: This fix should be cherry-pick into 3.1.x and subsequently pulled into 3.2.x.


bvenners:
I cherry picked these two commits into 3.1.x also.

PR #1388, opened 2018-07-02
Fix Retries.withRetry delay

marcuslonnberg:
Noticed that `Retries.withRetry` where not using its delay parameter. And that `Retries.withRetryOnCancel` where calling sleep twice.


plinlor:
@cla-bot[bot] check

artimasites:
@cla-bot[bot] check

PR #1383, opened 2018-06-19
Scala 2.13.0-M4 Support

cheeseng:
Initial version to support releasing for scala 2.13.0-M4.

PR #1369, opened 2018-05-18
Bump scalacheck version to 1.14.0

guersam:
Not using new features, just version bump due to the binary incompatibility.

PR #1389, opened 2018-07-05
try update scalacheck

kailuowang:


nigredo-tori:
This is already happening as a part of a much larger feature (#1383). The previous MR for this was closed for that reason (#1369).

kailuowang:
Thanks for the info @nigredo-tori

PR #1394, opened 2018-07-30
Fix Travis build with openjdk6

ashawley:
There are two issues with the Travis build:

- There's a hack required to get openjdk6 to work on Travis, see for example https://github.com/scala/scala-xml/pull/246
- There's a new  hack required to let sbt retrieve artifacts while using openjdk6 on Travis, which see https://github.com/scala/scala-xml/pull/247

However, if you are publishing from Travis or even openjdk6, you may have issues.  See discussion at https://github.com/scala/sbt-scala-module/issues/41

It seems like there are 3 places where `SBT_OPTS` are set and overwritten in the Scalatest build config.

- In `sbt_args` in `.travis.yml`
- In `SBT_OPTS` at the top of `travis_build.sh`
- And in each of the `MODE` builds of `travis_build.sh`

I'd like to just override them once when the JDK6 build is happening...

Fixes #1380.

PR #1400, opened 2018-08-15
Fix Travis build with openjdk6

cheeseng:
Changes is ported from: 

https://github.com/scalatest/scalatest/pull/1394

### DIFF
--- a/.sbtrepos
+++ b/.sbtrepos
@@ -1,0 +1,8 @@
+[repositories]
+  local
+  local-preloaded-ivy: file:///${sbt.preloaded-${sbt.global.base-${user.home}/.sbt}/preloaded/}, [organization]/[module]/[revision]/[type]s/[artifact](-[classifier]).[ext]
+  local-preloaded: file:///${sbt.preloaded-${sbt.global.base-${user.home}/.sbt}/preloaded/}
+  maven-central:  http://repo1.maven.org/maven2/
+  sonatype-public: http://oss.sonatype.org/content/repositories/public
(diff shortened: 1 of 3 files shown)


ashawley:
> Attempting to fix examples and examplesJS build failure in travis.

There were problems with the download from Ubuntu mirrors.  It didn't look it was the Scalatest build itself.  You should be able to re-run individual failing builds with the "rebuild" button

cheeseng:
@ashawley I see, good to know, anyway I think it is good idea to make examples and examplesJS to be consistent like the other, which the last commit does, let's see if it'll build successfully this time.  :)

cheeseng:
@ashawley By the way, I am looking at the current travis CI supported environments: 

https://docs.travis-ci.com/user/reference/overview/#virtualisation-environment-vs-operating-system

With 7.5G max memory may be we no longer need to break our tests in smaller builds that makes up the big build matrix on travis, I may try to do just scalatestTest(JS), gentests and examples(JS), probably in a separate PR from this one.

Do you think that's a good or bad idea?

ashawley:
I thought the build was broken up because of time limits.

https://docs.travis-ci.com/user/customizing-the-build/#build-timeouts

ashawley:
The 3.1 branch takes over 12 hours on Travis for 4 versions of Scala (2.10, 2.11, 2.12, and 2.13), so that's about 3 hours each.  If you tried to make just 3 builds, you'd be too close to the limit of 50 min.

ashawley:
One other benefit in splitting up the build in Travis is shorter build logs.  If we switched to running the build in larger chunks, the build logs get larger and become unreasonable and harder to look at in Travis.

PR #1407, opened 2018-08-21
Fix exit code in Travis script

ashawley:
In #1341, I got carried away in da481fe and didn't realize the script didn't have the `set -e` command so the exit status would get lost.  I saw some evidence that failing tests were improperly showing success.  We need to put back `exit $?` everywhere.

Looks like #1341 may have been forwarded to the 3.2.x branch, as well.


cheeseng:
@bvenners This PR looks good to me, please kindly review when you get a chance.

Thank you!

PR #1438, opened 2018-11-23
add _root_ to the macros defined in TypeMatcherMacro

bbonanno:
Sanitises the `TypeMatcherMacro` to avoid the following error when a val in scope is called `org` 

> Error:(180, 16) value scalatest is not a member of Organization 
         org.name shouldBe a[String]

which would be throw while compiling something like 

```scala
    "work when a val called 'org' is in scope" in {
      case class Organization(name: String)

      val org = Organization("acme")

      org.name shouldBe a[String]
    }
```

### DIFF
--- a/scalatest-test/src/test/scala/org/scalatest/matchers/TypeMatcherMacroSpec.scala
+++ b/scalatest-test/src/test/scala/org/scalatest/matchers/TypeMatcherMacroSpec.scala
@@ -1,0 +1,54 @@
+package org.scalatest.matchers
+
+import org.scalatest.{FlatSpec, Matchers}
+
+class TypeMatcherMacroSpec extends FlatSpec with Matchers {
+
+  case class Organization(name: String)
+  val org = Organization("acme")
+
+  "assertTypeImpl" should "work when a val called 'org' is in scope" in {
+    org.name shouldBe a[String]
+    org.name shouldBe an[String]
+  }
+
+  "aTypeMatcherImpl" should "work when a val called 'org' is in scope" in {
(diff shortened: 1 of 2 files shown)


cheeseng:
@bbonanno I am really sorry for attending really late for this, the fix looks good and it would be even better if we can add the test (like in the example) into our test so that we won't break it in the future, and I think we should include this fix in the upcoming 3.0.8 and cherry-pick this into 3.1.x.

Also, did you get any email about a CLA from us?

Thanks.

bbonanno:
@cheeseng no worries
I've added the tests for each one of the macros, it's a compile problem so the tests themselves are rather silly

Haven't got any email, what's the CLA about?

Cheers

cheeseng:
@bbonanno Sorry, it is about the following: 

https://www.artima.com/cla/contributorsLicenseAgreement.pdf

Do you mind to print, sign, scan and email back to cla@artima.com?

We apologize for any inconvenience caused, we have been wanting to automate it but unfortunately we're just not there yet.

bbonanno:
@cheeseng sent

cheeseng:
@bbonanno Great, thanks!

@bvenners The changes looks good to me, do you mind to review and merge it to 3.0.x branch?  Imho we can include it as part of the upcoming 3.0.8. 

Cheers.

bvenners:
I also cherry-picked this into 3.1.x: https://github.com/scalatest/scalatest/commit/0a0efb3df45afa84aed28ab8358f1df8787aadea

Issue #994, opened 2016-10-22
Discuss using `Await.result` in `ScalaFutures.futureResult`


thiloplanz:
I was wondering why `futureResult` uses polling instead of relying on `Await.result` provided by the Scala standard library.

[Other people]( http://cjwebb.github.io/blog/2015/02/02/scalatest-futures/) seem to have the same question:  

> Strangely, ScalaTest chooses to poll futures, despite both Scala and Twitter Futures coming with Await functions that handle timeouts. Using that as a starting point would have seemed more sensible to me.

Looking at the source (`ScalaFutures.scala`), there is already a method that seems to do something like that (it calls `Await.ready`), but it is commented out. The revision log does not shed light on the history there.

Can you leave some comment in the source code as to why `Await.result` cannot be used (or if it can be used, restore the method)?

fsvehla:
I’d love if it would use `Await.result` when it can, currently I have to provide my own extension method in all projects that calls `Await.result` with the implicit `PatienceConfig`’s timeout

henrikno:
I came here to ask the exact same question. Git history doesn't give any clues.

bvenners:
@plinlor see if we can use Await.result instead of polling.

jducoeur:
Does `futureResult` need to work in Scala.js?  If so, `Await.result` isn't an option -- blocking is flatly impossible in the SJS environment.  That might explain why it got commented out...

henrikno:
Is it possible to implement Async.result as polling in scalajs? Or is it possible to have a check for scalajs and choose implementation based on that?

jducoeur:
Likely possible by smart use of crossProject, but I don't know this code,
so can't guess how much the effort would be...

plinlor:
@cheeseng Do you agree that Await.result won't work on Scala.js?

jducoeur:
Seriously, this one's in my area of expertise -- I'm one of the most experienced Scala.js devs out there.  Await.result is *impossible* on Scala.js: it doesn't exist and likely never will, because of how JavaScript operates...

durban:
Using `Await.result` (or at least `scala.concurrent.blocking` around the `Thread.sleep` call) on the JVM would also have the advantage of having less chance to deadlock (if the used `ExecutionContext` supports managed blocking).

kwinter:
[...]

PR #1457, opened 2018-12-21
add blocking context to SleepHelper

kwinter:
while it's not Await.result, should help with issues mentioned in #994


kwinter:
closing to re-open and trigger a new CI build

PR #1462, opened 2018-12-28
Change packages of mocking and Java testing to scalatestplus

cheeseng:
Changed these package names:

org.scalatest.easymock => org.scalatestplus.easymock
org.scalatest.jmock => org.scalatestplus.jmock
org.scalatest.mockito => org.scalatestplus.mockito
org.scalatest.selenium => org.scalatestplus.selenium
org.scalatest.junit => org.scalatestplus.junit
org.scalatest.testng => org.scalatestplus.testng

and left deprecated aliases in the original packages.


scalatest/src/main/scala/org/scalatest/Suite.scala
```diff
@@ -2170,6 +2147,30 @@ used for test events like succeeded/failed, etc.
     (Map[A, B]() /: (for (m <- ms; kv <- m) yield kv)) { (a, kv) =>
       a + (if (a.contains(kv._1)) kv._1 -> f(a(kv._1), kv._2) else kv)
     }
+
+  // SKIP-SCALATESTJS-START
+  // Leave this around for a while so can print out a warning if we find testXXX methods.
+  def yeOldeTestNames(theSuite: Suite): Set[String] = {
```
bvenners: I think we can get rid of this now. It has been "a while" since the 2.0 release where we stopped supporting test methods on Suite.

scalatest/src/main/scala/org/scalatestplus/junit/AssertionsForJUnit.scala
```diff
@@ -96,7 +95,7 @@ import org.scalactic._
  */
 trait AssertionsForJUnit extends Assertions {
 
-  private[scalatest] override def newAssertionFailedException(optionalMessage: Option[String], optionalCause: Option[Throwable], pos: source.Position): Throwable = {
+  private[org] override def newAssertionFailedException(optionalMessage: Option[String], optionalCause: Option[Throwable], pos: source.Position): Throwable = {
     new JUnitTestFailedError(optionalMessage, optionalCause, pos, None)
```
bvenners: Instead of private[org] we should make a copy of these and move them to the scalatestplus.* packages where they are needed. Eventually the stuff in the scalatestplus.* packages will go to a separate github repo.

scalatest/src/main/scala/org/scalatestplus/selenium/WebBrowser.scala
```diff
@@ -45,10 +45,9 @@ import org.openqa.selenium.support.ui.Select
 import org.scalatest.exceptions.TestFailedException
 import org.scalatest.exceptions.StackDepthException
 import org.openqa.selenium.JavascriptExecutor
-import org.scalatest.ScreenshotCapturer
+import org.scalatest.ScalaTestInternals.ScreenshotCapturer
 import org.scalatest.time.Nanosecond
```
bvenners: @cheeseng Shouldn't ScreenshotCapturer just be in org.scalatestplus.selenium? Is it used elsewhere?

scalatest/src/main/scala/org/scalatestplus/testng/TestNGSuiteLike.scala
```diff
@@ -13,19 +13,21 @@
...
-import Suite.getIndentedTextForTest
-import Suite.wrapReporterIfNecessary
+import org.scalatest.ScalaTestInternals.Suite
+import org.scalatest.ScalaTestInternals.Suite.formatterForSuiteAborted
+import org.scalatest.ScalaTestInternals.Suite.getIndentedTextForTest
+import org.scalatest.ScalaTestInternals.Suite.wrapReporterIfNecessary
+import org.scalatest.ScalaTestInternals.Suite.yeOldeTestNames
+import org.scalatest.ScalaTestInternals.ScalaTestStatefulStatus
+import org.scalatest.ScalaTestInternals.Resources
 import events.MotionToSuppress
```
bvenners: We should copy these guys over to the org.scalatestplus.testng package and get rid of ScalaTestInternals, which is private[org].

bvenners:
I've merged most of the commits into 3.0.x, all but the last four that brought in the big refactor. I'll close this one and will look at what alternatives we have. I think we can make something public or protected instead in this case. The reason is that run is not just a method we call on Suite, but is available for all. So who knows who has called it and done what. It was there so on purpose to be called, so that's not a we need to be careful not to break that. Moving the wrapReporterIfNecessary I worry might break something somewhere, so in this case it is probably better to just bring that forward into the published API.

PR #1464, opened 2019-01-04
Remove a no-longer-relevant TODO

jducoeur:
It looks to me like it's already done.


jducoeur:
Withdrawing this PR, since I'm about to do much more work in the same file, and it's easier to deal with it all together.

PR #1476, opened 2019-01-09
Reworked most nextXXXFloat and nextXXXDouble in Randomizer

bvenners:
I did not look at the between 0.0 and 1.0 float and double methods yet, nor did I do anything with the choose methods.

I did get one flicker when running the tests:

```
[info] CommonGeneratorsSpec:

[info] The CommonGenerators object should offer a frequency method that takes a varargs of Int weights to generators and produces a generator that 

[info] - returns values from each specific generator with a probability determined by the weights *** FAILED *** (88 milliseconds)

[info]   59 was not 57 plus or minus 1 (CommonGeneratorsSpec.scala:3169)
```

This test just works with Ints, which I didn't change in this PR. So it is possible it was due to some change in Justin's PR or just a plain old flicker that happened to reappear. I didn't look at it closely, but will do so tomorrow.


scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
```diff
@@ -149,11 +150,19 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
...
-    (java.lang.Float.intBitsToFloat((s << 31) | (e << 23) | m), rm)
+  def nextFloat: (Float, Randomizer) = {
+    // The space of possible Floats
+    // 2 to the power of 32, which is covered by Int.MinValue to Int.MaxValue
+    // - (2 to the power of 23 (23 is the number of mantissa bits))
+    // + (3 for the two infinities and one NaN)
+
+    // scala> Int.MaxValue - math.pow(2, 23).toInt + 2
+    // res13: Int = 2139095042
+    val (x, r) = chooseInt(Int.MinValue, 2139095042)
```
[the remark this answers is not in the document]
jducoeur: Also, IMO this is being more precise than is really warranted (and expending more effort) to very slightly improve the distribution of an extreme edge case.  I'd actually recommend just calling `nextInt()` here -- it's faster, and plenty good enough.

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
```diff
@@ -149,11 +150,19 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
...
+    // - (2 to the power of 23 (23 is the number of mantissa bits))
+    // + (3 for the two infinities and one NaN)
+
+    // scala> Int.MaxValue - math.pow(2, 23).toInt + 2
+    // res13: Int = 2139095042
+    val (x, r) = chooseInt(Int.MinValue, 2139095042) 
+
+    // Pick one lucky number to play the lotto with:
+    if (x == 999) (Float.NaN, r)
+    else r.nextExtRealFloatValue
```
jducoeur: Mostly at an aesthetic level, I'm a little unhappy about the inefficiency of the new algorithm -- especially when you factor in what `nextExtRealFloatValue` does, we're generating two random numbers solely for the fantasmagorically small chance that they will result in one of your magic numbers.  It's minor, but it's wasted cycles in virtually all cases.

Can/should we compose these operations somehow?  This feels like there might be a better factoring...

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
```diff
@@ -179,10 +191,19 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
...
-    val (m, rm) = re.chooseLong(0L, 0xfffffffffffffL)
-    (java.lang.Double.longBitsToDouble((s << 63) | (e << 52) | m), rm)
+    // The space of possible Doubles
+    // 2 to the power of 64, which is covered by Long.MinValue to Long.MaxValue
+    // - (2 to the power of 52 (52 is the number of mantissa bits))
+    // + (3 for the two infinities and one NaN)
+
+    // scala> Long.MaxValue - math.pow(2, 52).toLong + 3
+    // res27: Long = 9218868437227405314
+    val (x, r) = chooseLong(Long.MinValue, 9218868437227405314L)
```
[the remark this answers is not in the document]
jducoeur: (And same comment that `nextLong` is probably good enough, and should probably be used instead.)

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
```diff
@@ -166,11 +175,14 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
     * @return A random Double in that range, and the next Randomizer to use.
     */
   def nextDoubleBetween0And1: (Double, Randomizer) = {
+    // TODO: Verify this implementation against the spec and comment it.
     val (ia, ra) = thisRandomizer.next(26)
     val (ib, rb) = ra.next(27)
     (((ia.toLong << 27) + ib) / (1L << 53).toDouble, rb)
   }
 
+   // TODO: Any reason why we have both a next and a nextInt? Should we drop next?
```
jducoeur: I don't understand -- `next()` is the method that underpins all the `nextXX()` functions.  How would you drop it?

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
```diff
@@ -280,47 +328,113 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
...
+    // res30: Float = Infinity
+    //
+    // scala> -0.0f.abs
+    // res31: Float = 0.0
+    //
+    // scala> PosZFloat(-0.0f)
+    // res0: org.scalactic.anyvals.PosZFloat = PosZFloat(-0.0f)
+    val (candidate, r) = nextExtRealFloatValue
+    val posZ =
+      if (isNegativeZeroFloat(candidate)) candidate // leave it negative zero
```
jducoeur: In principle, this seems wrong.  Granted, the distinction between negative and positive zero is at the angels-dancing-on-a-pin level, but it feels inappropriate for PosZFloat to include negative zero; I would intuitively expect this to become positive zero...

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
```diff
@@ -280,47 +328,113 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
...
-        case Double.PositiveInfinity => Double.MaxValue
-        case Double.NegativeInfinity => Double.MaxValue
-        case _ => n
-      }
-    (FiniteDouble.ensuringValid(finite), r)
+    // The exponent portion of a Double occupies 11 bits. It can be 0 (which represents an exponent of -1023)
+    // to 2047 (which is a reserved value for NaN values and +/- infinity). The highest regular (non-reserved)
+    // exponent therefore is 2046, which in hex is 0x7fe (which represents an exponent of +1023).
+    // Thus by chosing the exponent Int between 0 and 0x7fe, we can't get a NaN or an infinity.
+    val (s, rs) = chooseLong(0L, 1L)                  // The sign bit (1 bit)
```
jducoeur: Is `chooseLong()` the right tool for this job?  Seems like an awful lot of cycles to choose between 0 and 1.  We might want a simple `nextBit()` for this.

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
```diff
@@ -280,47 +328,113 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
...
+  private def nextExtRealDoubleValue: (Double, Randomizer) = {
+
+    // The space of possible non-NaN Doubles
+    // 2 to the power of 64, which is covered by Long.MinValue to Long.MaxValue
+    // - (2 to the power of 52 (52 is the number of mantissa bits))
+    // + (2 for the two infinities)
+
+    // scala> Long.MaxValue - math.pow(2, 52).toLong + 2
+    // res27: Long = 9218868437227405313
+    val (x, r) = chooseLong(Long.MinValue, 9218868437227405313L)
```
jducoeur: Same comment as before: I'd write this in hex.

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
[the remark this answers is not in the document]
bvenners: I'll switch that to hex and think about using nextInt.

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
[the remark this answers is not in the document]
bvenners: Will change to Hex and think about nextLong. These are way less likely to happen then getting struck by lightening, so question is if we're going to do something that will never happen, does it matter if we do it precisely?

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
bvenners: I'll think about this when I look at our chooseInt algos, which is coming up.

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
bvenners: Yes I realized next takes bits and nextInt does not, so I deleted this comment in my later PR.

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
bvenners: I know, but I didn't think of this before releasing PosZFloat and PosZDouble to the wild, and they allow -0.0. So changing that now would need to go through a deprecation cycle with renaming, which would be a pain. But also -0.0 is actually a Float zero. So it could go either way. PosInt means the postitive part of the space, PosZInt means the positive part of the space plus zero. It doesn't mean "positive zero," because there is no such thing. But let's keep thinking about it. It was surprising to me, and there is at least one way I know of that the -0.0. can be transformed to a non-zero value but still be negative:

```
scala> 1 / -0.0
res7: Double = -Infinity
```

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
bvenners: I'll look at next(1).

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
bvenners: Will do.

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
```diff
@@ -149,11 +150,19 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
...
+    // 2 to the power of 32, which is covered by Int.MinValue to Int.MaxValue
+    // - (2 to the power of 23 (23 is the number of mantissa bits))
+    // + (3 for the two infinities and one NaN)
+
+    // scala> Int.MaxValue - math.pow(2, 23).toInt + 2
+    // res13: Int = 2139095042
+    val (x, r) = chooseInt(Int.MinValue, 2139095042) 
+
+    // Pick one lucky number to play the lotto with:
+    if (x == 999) (Float.NaN, r)
```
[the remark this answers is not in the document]
jducoeur: While I can understand that, I think you're introducing a real problem here: you've built a better mousetrap, and expecting folks to only use it in one room is probably a mistake.

That is, what you've got here, intentionally, is a much better replacement for `java.util.Random`.  Specifically, you have a replacement that is designed to work hand-in-glove with Scalactic and its types.

That being the case, I would expect anybody *using* Scalactic who is using randomness for *anything* to wind up using this class.  Saying "It is not intended to be used in production" is probably futile: folks sometimes *do* use randomness in production code, and you've got a rich randomness library, so I expect people to use it.

I suppose what I'm arguing here is that Randomizer is in the wrong place.  It probably belongs in Scalactic, not in ScalaTest, and should be considered with that in mind.  You've built a more general tool than I think you may have intended, and I suspect folks will use it as such...

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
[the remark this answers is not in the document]
jducoeur: Well, that's my argument for `nextLong` -- it's strictly more efficient, so I'm arguing that it's appropriate not to waste cycles on making something extremely unlikely slightly more precise.

(If this happened *frequently*, I might sympathize more with spending the CPU on getting it exactly right.)

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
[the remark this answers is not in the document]
bvenners: Well, gnerating random numbers for testing is a different use case than the many different use cases for random numbers someone might want in production. These aren't secure. In production you may or may not want security. They favor touching the whole space of possibility over raw speed. The replacement nextFloat/DoubleBetween0And1 algo is probably slower than the one used in java.util.Random, but touches all the points in the space of possibilities, whereas the one in java.util.Random does not. This Randomizer generates undesirable values like infinities and NaNs, which are useful in testing, but not in production. This Randomizer requires that you pass the thing through each time in a functional way so that users can get repeatable test results. That's nicely functional, so maybe people want that in production too, but that's different from java.util.Random.

If you feel the urge to make random number generator that knows about Scalactic anyvals that's a different problem. I have never heard of that need, so I don't know anything about what people might want. It seems plausible, but that's a different design for a different day.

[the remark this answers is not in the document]
jducoeur: Well, my final word on the subject is, at least make the ScalaDoc more explicit about this.  It's not a topic I was thinking about when I was documenting `nextFloat()` and `nextDouble()` (didn't even occur to me), but if those can return NaNs and infinities, it should be crystal-clear that it's going to do so, because otherwise I think that many (I suspect most) folks won't even consider that possibility.  It's a driving reason why they may want to seek out other functions instead...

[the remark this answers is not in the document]
bvenners: For sure. I noticed a few spots where we need to add more information, such as saying the Between methods are inclusive of 0 but exclusive of 1. I'll do a pass over the docs before we release.

scalatest-test/src/test/scala/org/scalatest/prop/RandomizerSpec.scala
```diff
@@ -53,38 +53,6 @@ class RandomizerSpec extends FunSpec with Matchers {
       jc shouldEqual ic
     }
 
-    it("should offer a nextFloatBetween0And1 method that produces the same values as java.util.Random given the same seed") {
```
cheeseng: @bvenners Since the old tests are being removed, I think it is good to add back a test that make sure the method returns the value between 0.0-1.0, e.g. may just generate 10 numbers and make sure they are in the range.

PR #1477, opened 2019-01-10
Improve edges composition

bvenners:
Hi Chee Seng & Justin,

This fixes a bug in the flatMap method's algorithm for composing edges so that it short circuits as intended once the requested maximum number of edges is obtained. It also enhances the algorithm (using some list shuffles) so that when combinations of edges exceed the max requested length, we now randomly shuffle the data being composed so that each time we run our short circuiting will happen on different points of the space of combinations.

The goal is that over time we'll touch all the combinations of edges. But I'm not 100% sure it will achieve that. It would be good if you could study the algo and see if that makes sense, or does not, if you can think of some other approach that could achieve that goal.

Thanks.

Bill

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/prop/Generator.scala
+++ b/scalatest/src/main/scala/org/scalatest/prop/Generator.scala
@@ -87,7 +87,6 @@ trait Generator[T] { thisGeneratorOfT =>
               remainingGenOfU match {
                 case head :: tail =>
                   val (listOfU, nnRnd) = head.initEdges(maxLength, nRnd)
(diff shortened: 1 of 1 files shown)


jducoeur:
How soon do you need me to review this?  I suspect it'll be next week before I'm sufficiently read into Generator to be able to do so in an informed way...

bvenners:
@jducoeur Well i'd like to merge it, so we can let Chee Seng merge it. When you get to Generator you will review all the code anyway, including this bit. This is needed before adding edges, which I'd like to do after I finish up the Randomizer stuff.

PR #1478, opened 2019-01-10
Replace algos for floats and doubles between 0 and 1 in Randomizer

bvenners:
This should go in after #1476.

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
+++ b/scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
@@ -138,9 +138,48 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
     * @return A random Float in that range, and the next Randomizer to use.
     */
   def nextFloatBetween0And1: (Float, Randomizer) = {
-    // TODO: Verify this algo against the spec, and comment it.
+    /*
+    0.0f is 32 zero bits. The sign is 0 for positive. The exponent of 0 is a special
+    case for 0.0 and for subnormal values. It means 0 if the mantissa is also 0.
+    1.0 is represented as 1.00 * (2 ** 0). The 1 to the left of the binary point
+    is implicit, so this means that the mantissa is all zeros again for 1.0. An
+    exponent of 0 is represented by 127 (0x7f). (To get the actual exponent from the biased
+    one, subtract 127.)
+
+    So except for 0.0f and 1.0f, which we'll need to handle specially, we can generate
+    a random 23 bits for the manitissa, and for the exponent a number between 1 and 126 (0x7e).
+    */
+
+    // The total number of possible images for Float between 0.0 and 1.0 is:
+    // (2 ** 23) * (126) + 2
+    // 2 to the power of 32 (0x800000) is how many different mantissas can be represented in 23 bits.
+    // 126 (or 0x7e) is how many exponents there are for numbers between 0.0 and 1.0, exclusive.
+    // 2 is for 0.0 and 1.0, which we must handle specially, because their mantissas must be all zeros.
+    // scala> 0x800000 * 0x7e + 2
+    // res14: Int = 1056964610
+    val (x, r) = chooseInt(0, 1056964610)
+
+    // Pick two lucky numbers to play the lotto with:
+    if (x == 333)
+      (0.0f, r)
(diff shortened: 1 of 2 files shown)


cheeseng:
@bvenners @jducoeur  Imho the java version is simpler, it is simply: 

(choose a random between min and max value) / max value

e.g. if the the possible value range is 0 -100, then it is: 

(choose a random between 0 - 100) / 100

which will return a value between 0.00 - 1.00

In the case of float, as in the javadoc: 

return next(24) / ((float)(1 << 24));

If I understand it correctly it means: 

(choose a random between 0 - 2^24, inclusive) / (2^24).toFloat
= 
(choose a random between 0 - 16777216) / 16777216.0

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
```diff
@@ -138,8 +138,38 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
...
+    */
+
+    // The total number of possible images for Float between 0.0 and 1.0 is:
+    // (2 ** 23) * (126) + 2
+    // 2 to the power of 23 (0x800000) is how many different mantissas can be represented in 23 bits.
+    // 126 (or 0x7e) is how many exponents there are for numbers between 0.0f and 1.0f, exclusive.
+    // 2 is for 0.0f and 1.0f, which we must handle specially, because their mantissas must be all zeros.
+    // scala> 0x800000 * 0x7e + 2
+    // res14: Int = 1056964610
+    val (x, r) = chooseInt(0, 1056964610)
```
jducoeur: Same comment as the other review: it's a bit-centric number, so I'd write this in hex.

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
```diff
@@ -138,8 +138,38 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
...
+
+    // Pick two lucky numbers to play the lotto with:
+    if (x == 333)
+      (0.0f, r)
+    else if (x == 222)
+      (1.0f, r)
```
jducoeur: This is inclusive of 1?  That's actually not what I would expect.  Have you checked that the various uses are comfortable with it?

(It's not obviously wrong, just unintuitive to me.  I would recommend tweaking the ScalaDoc to mention that explicitly.)

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
```diff
@@ -166,11 +204,57 @@ class Randomizer(private[scalatest] val seed: Long) { thisRandomizer =>
...
+
+    // The total number of possible images for Double between 0.0 and 1.0 is:
+    // (2 ** 52) * (1022) + 2
+    // 2 to the power of 52 (0x10000000000000L) is how many different mantissas can be represented in 52 bits.
+    // 1022 (or 0x3fe) is how many exponents there are for numbers between 0.0 and 1.0, exclusive.
+    // 2 is for 0.0 and 1.0, which we must handle specially, because their mantissas must be all zeros.
+
+    // scala> 0x10000000000000L * 0x3fe + 2
+    // res20: Long = 4602678819172646914
+    val (x, r) = chooseLong(0, 4602678819172646914L)
```
jducoeur: Usual comment about hex.

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
bvenners: Will make it hex.

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
bvenners: Oops that is a bug. It should indeed be exclusive of 1.0. I'll fix that.

scalatest/src/main/scala/org/scalatest/prop/Randomizer.scala
bvenners: Will hexify.

bvenners:
I'm going to close this one and pull these changes into the branch being used by PR 1476, make the review suggestions there, and then we can merge all of that in via 1476.

PR #1482, opened 2019-01-12
Fixing Travis Build for 3.0.x

cheeseng:
Trying to fix jdk6 travis build problem by switching to precise.


cheeseng:
Hi @ashawley , thanks for pointing that out, the current travis seems to be broken due to jdk6, we're considering to drop support for jdk6 in 3.1.x, but in 3.0.x we still need to support jdk6 though (the purpose of this PR).  Anyway, PR #1407 seems to be a good one to be merged though.

ashawley:
Yeah, it won't help with fixing your JDK6 issues, I just mentioned it in the context of getting Travis working.

I'll try to get my JDK6 tricks for Travis and see if I can share them with you for ScalaTest.  One piece is https://github.com/scala/scala-xml/pull/247, but whether it works depends on how much customization exists in the ScalaTest build.

ashawley:
Oh, looks like we added the sbt tricks for JDK 6 back in #1400.  I forgot.  

I think you fixed Travis in ee736db.  The problem is you need to fix `genColCompatTests`.  It's only failing because it is overwriting `SBT_OPTS`.  You need to remove the `export SBT_OPTS` here https://github.com/scalatest/scalatest/blob/4f70f46c3d/travis_build.sh#L284

ashawley:
Looks like it worked!

project/plugins.sbt
```diff
@@ -1,3 +1,5 @@
+resolvers += "typesafe" at "http://repo.typesafe.com/typesafe/releases"
+
```
ashawley: This doesn't do anything, so you can remove it.

cheeseng: @ashawley Yes, thank you very much for the fix suggestion about the extra SBT_OPTS, I was puzzling what's went wrong and you saved my day!

And yes the typesafe repo thing above didn't do anything, I'll remove it and after it builds green again I'll remove the [DO-NOT-MERGE] tag for this PR.

Thank you again for the fix!

PR #1483, opened 2019-01-14
[DNM] First-pass documentation for PropertyArgument and PropertyCheckResult.

jducoeur:
This is *not* complete yet -- it needs to be finished once I understand some of the details better, especially WRT how PropertyArgument is actually used in practice. But it's a decent start.

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/prop/PropertyCheckResult.scala
+++ b/scalatest/src/main/scala/org/scalatest/prop/PropertyCheckResult.scala
@@ -15,14 +15,72 @@
  */
 package org.scalatest.prop
 
+/**
+  * Describes the outcome of a Property Check operation such as 'forAll()'.
+  *
+  * This will always be one of three subtypes: [[org.scalatest.prop.PropertyCheckResult.Success]],
+  * [[org.scalatest.prop.PropertyCheckResult.Exhausted]], or [[org.scalatest.prop.PropertyCheckResult.Failure]].
+  * See those subclasses for details on what each one means.
+  */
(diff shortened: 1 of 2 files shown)


scalatest/src/main/scala/org/scalatest/prop/PropertyCheckResult.scala
```diff
@@ -15,14 +15,72 @@
...
+    *
+    * This generally indicates a bug in your code. Values were generated that turned out to produce an error.
+    * The next step is usually to examine these values, and understand why they caused an error.
+    *
+    * Note that this does not ``necessarily`` mean that the code under test is in error: it is possible that the
+    * property itself is incorrect. It is good practice to check both, and figure out which one is wrong.
+    *
+    * @param succeeded How many generated values passed the property before the failure.
+    * @param ex The Exception that was thrown by the property, if any.
+    * @param names TODO: how does this parameter differ from argsPassed?
```
bvenners: I don't remember the difference between names and argsPassed.label. I suspect I was trying to make the error message nice by letting users see the variable name in the source code in the error message itself. That would require a macro, and I'm not sure we are doing that. So we'll need to do more digging to figure that out.

jducoeur: Okay, cool.  For now, I'm leaving the TODO as a string around my finger, to remember to keep an eye out for this as I'm digging around in the rest of the code.  (I did a quick scan, but didn't get far enough yet to be clear about usage -- the call stacks are pretty complex.)

Issue #1487, opened 2019-01-17
PatienceConfig should not be a path-dependant type


Woodz:
Currently, `PatienceConfig` is a path-dependant type as it is declared within `AbstractPatienceConfiguration.scala`. This means that there are different types of `PatienceConfig` depending on the inheritance/import hierarchy.

This causes problems when using both `ScalaFutures` and `Eventually` in the same test suite, as there are different implicit values for both `ScalaFutures.PatienceConfig` and `Eventually.PatienceConfig`. This causes issues at execution time due to timeouts not being respected, which is very hard to diagnose.

Minimal reproducible:
```
import org.scalatest.WordSpec
import org.scalatest.concurrent.{Eventually, IntegrationPatience}
import org.scalatest.concurrent.ScalaFutures._

class PatienceConfigTest extends WordSpec with Eventually with IntegrationPatience {
  // Which one of these has the integration patience applied?
  val implicitScalaFuturesPatienceConfig: ScalaFutures.PatienceConfig = implicitly[ScalaFutures.PatienceConfig]
  val implicitEventuallyPatienceConfig: Eventually.PatienceConfig = implicitly[Eventually.PatienceConfig]
}
```

PR #1488, opened 2019-01-17
Move `PatienceConfig` out of `AbstractPatienceConfiguration`

Woodz:
Change `PatienceConfig` to no longer be a path-dependant type to avoid
 confusion with multiple instances.
Add a protected `defaultPatienceConfig` to
 `AbstractPatienceConfiguration` to allow scaling and overriding by
 derived traits
Resolve: #1487

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/concurrent/AbstractPatienceConfiguration.scala
+++ b/scalatest/src/main/scala/org/scalatest/concurrent/AbstractPatienceConfiguration.scala
@@ -53,42 +53,8 @@ import org.scalatest.time.{Span, Millis}
  */
 trait AbstractPatienceConfiguration extends ScaledTimeSpans {
 
-  /**
-   * Configuration object for asynchronous constructs, such as those provided by traits <a href="Eventually.html"><code>Eventually</code></a> and
-   * <a href="AsyncAssertions.html"><code>AsyncAssertions</code></a>.
-   *
-   * <p>
-   * The default values for the parameters are:
-   * </p>
-   *
-   * <table style="border-collapse: collapse; border: 1px solid black">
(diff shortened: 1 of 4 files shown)


Woodz:
I'm not sure which base branch I should use for this PR - GitHub is configured that 3.1.x is the default branch but this does not compile for me due to missing Scala 2.12 native-scala dependencies, so I thought master would be sensible as there is not a documented branching strategy for this project that I could find

bvenners:
@Woodz can you give an example where you had trouble with confusion of multiple instances?

Woodz:
> @Woodz can you give an example where you had trouble with confusion of multiple instances?

Sure! The scenario is documented in issue 1487, which is a combination of using `ScalaFutures` companion object with `Eventually` trait plus `IntegrationPatience`. This end result is that the `PatienceConfig` used by `futureValue` is different from the one increased by `IntegrationPatience`. Let me know whether this is unclear

Woodz:
Also, please note that due to https://github.com/scalatest/scalatest/issues/1492 I am unable to run the tests on my machine so cannot tell whether I have broken any tests, so would appreciate if this could either be manually executed or run through Travis to confirm

devzer01:
why hasn't this been merged yet?

plinlor:
@cla-bot[bot] check

artimasites:
@cla-bot[bot] check

Woodz:
> Hi @Woodz, we require contributors to sign our Contributor License Agreement, and we don't have yours on file. In order for us to review and merge your code, please access https://www.artima.com/cla/choose-type to sign our Contributor License Agreement. Your effort is highly appreciated. Thank you.

Signed. Apologies for the delay

PR #1489, opened 2019-01-17
SBT 1.x Build

cheeseng:
Updated to use SBT 1.1.4, path the way to include Dotty build.


cheeseng:
@bvenners Travis passed, this is ready for your review now.

PR #1506, opened 2019-01-31
Better Fix for Init Seed

cheeseng:
A different way to fix init seed problem, this approach won't have confusing init seed in forAll's error message problem even when a fixed Randomizer (created from apply method) is used with a custom Generator passed to the forAll function.


jducoeur:
I like this -- it's similar to one of the approaches I was considering, but I think this one's better.  It's a more general fix than the previous one, and achieves what I'd really like to see, that calls to `Randomizer(123)` can see `123` as the `seed`.

Bill, when you get a chance, I'd recommend seriously considering this tweak...

Issue #1500, opened 2019-01-24
Please add a Chooser typeclass to prop in 3.1.x


bvenners:
Hi Justin,

Per our discussion today please add a Chooser[T] typeclass to the prop package, and use it in CommonGenerators.valuesBetween like:

```
def valuesBetween[T](from: T, to: T)(implicit genOfT: Generator[T], ordOfT: Ordering[T], chooser: Choosable[T]): Generator[T]
```

We're thinking this will enable syntax like:

```
forAll (valuesBetween(Gizmo(3), Gizmo(5)) { (t: T) => … }
```

and:

```
val gizmosBetween = valuesBetween[Gizmo] _
```

then:

```
forAll (gizmosBetween(Gizmo(3), Gizmo(5)) { (t: T) => … }
```

For any Gizmo for which a Ordering[Gizmo] and a Generator[Gizmo] is defined.

Thanks.

Bill

jducoeur:
Bill, I believe this one is done -- want to close it?

PR #1510, opened 2019-02-04
Add Chooser typeclass

jducoeur:
This is the generalization of the `chooseXxs` functions in `Randomizer`.  It allows us to replace the previous, messy `valuesBetween()` function with a really elegant `between()` instead.

Includes instances for all of our usual numeric types.

Existing unit tests in CommonGeneratorsSpec already hammer these code paths hard (and, indeed, turned up a bug in between()), so I think this is properly tested.


jducoeur:
This fixes #1500

PR #1514, opened 2019-02-06
Finished the documentation for CommonGenerators

jducoeur:
Much of this is simply about filling in the boilerplate for the Generators that are just external references.

But I also decided that this list is long enough to warrant using Scaladoc's @group mechanism. (Note the tweaked to the build file to turn that on.) The order of the sections is arbitrary, and be be easily rearranged by changing the values on the @groupprio entries in the top-level Scaladoc.


bvenners:
The groupings look great on CommonGenerators.

Issue #1517, opened 2019-02-07
Please ensure underlying exception is displayed when a property-based test fails


bvenners:
Hi Justin,

In 3.1.x, when a property-based test fails, we don't see the underlying exception. If the user sets -oS or -oF, they do see the exception thrown by forall, but that one does not contain as a "cause" exception the assertion that failed within the forAll's body. I think the way to make this better is to set the cause of the exception currently being thrown by forAll to the one that the assertion actually failed with.

Thanks.

Bill

PR #1520, opened 2019-02-12
#1517: display the actual Exception when a property check fails

jducoeur:
The Exception being generated for this case had two distinct problems:

* First, it wasn't recording the inner Exception as the cause of the outer Exception. So the -oS and -oF flags weren't showing the inner stack.
* Second, the error message was hard-coded to use the outer Exception, and completely ignored the inner one. This needed some careful rewriting.

To make the problem more manageable, merged together the two (identical) clauses that were creating the error message, into a single common function. (In general, PropCheckAsserting desperately needs a refactor.)


bvenners:
Closing this because reworking it in this branch: https://github.com/scalatest/scalatest/tree/feature-implement-shrink-for-forall-with-one-arg

PR #1521, opened 2019-02-13
Feature add shrink and canonicals for nonzeroint

bvenners:


scalatest-test/src/test/scala/org/scalatest/prop/GeneratorSpec.scala
```diff
@@ -1722,6 +1722,34 @@ class GeneratorSpec extends FunSpec with Matchers {
         edges should contain (NonZeroInt(1))
         edges should contain (NonZeroInt.MinValue)
       }
+      it("should produce NonZeroInt canonical values") {
+        import Generator._
+        val gen = nonZeroIntGenerator
+        val (canonicals, _) = gen.canonicals(Randomizer.default)
+        canonicals.toList shouldBe List(NonZeroInt(1), NonZeroInt(-1), NonZeroInt(2), NonZeroInt(-2), NonZeroInt(3), NonZeroInt(-3))
```
jducoeur: On principle (minor point), I'd say that this should be testing sameElements, rather than testing the exact order of the List -- this is over-specified (which isn't good testing practice IMO), and doesn't account for the fact that nonZeroIntGenerator would still be entirely valid if it was using the Randomizer to reorder the elements.

scalatest-test/src/test/scala/org/scalatest/prop/GeneratorSpec.scala
```diff
@@ -1722,6 +1722,34 @@ class GeneratorSpec extends FunSpec with Matchers {
...
+      }
+      it("should shrink NonZeroInts by repeatedly halving and negating") {
```
jducoeur: Is it appropriate to specify this quite so tightly?  This feels like testing the implementation rather than the requirement.  Again not end of the world, but it feels too coupled to the implementation...

scalatest-test/src/test/scala/org/scalatest/prop/GeneratorSpec.scala
bvenners: yes, i agree. i'll push that change to this branch.

scalatest-test/src/test/scala/org/scalatest/prop/GeneratorSpec.scala
bvenners: Not sure. I think that's the spec in this case. Let's leave this one as is for now.

scalatest-test/src/test/scala/org/scalatest/prop/GeneratorSpec.scala
bvenners: Actually I think no, because this is used for the shrink on composed generators, and the items returned by shrink should be in increasing order. Then reason is that the first one that fails the test wins and we don't look any farther.

jducoeur: Hmm.  Okay, that's plausible -- but if it's true, then we need to add it to the documentation.  (If I didn't know that, it's certainly not going to be obvious to the typical user.)

bvenners: Yes, that should be in our documentation, the notion that shrink should return values in "increasing bigness" order.

bvenners: ...and that canonicals are used for shrink in composed generators, and therefore it also needs to be in increasing bigness order.

jducoeur: Right.  The docs on shrink already talk about this (actually, it's that all the values returned from shrink must be *smaller* than the parameter passed in), but the ones for canonicals don't.

Issue #1519, opened 2019-02-12
Please make generators for options, eithers, or ors


bvenners:
Hi Justin,

I noticed we are not providing generators for Options, Eithers, and Ors yet. Please add those to CommonGenerators. The names should probably be options, eithers, and ors, to be consistent.

Thanks.

Bill

PR #1522, opened 2019-02-14
#1519: add/complete Generators for Option, Or and Either

jducoeur:
There were partial Generators already for Option and Or, but they were only half-finished, and had no tests.

Did a little refactoring in CommonGeneratorsSpec, to reduce unnecessary boilerplate.  We might want to go through and apply that more broadly -- it would make the file *much* more tractable.

Discovered along the way that CommonGeneratorsSpec had a serious bug, such that most of the tests were doing much less than they appeared.  Fixed that.

That revealed that the function1/2s tests were fundamentally broken -- they were trying an impossible test, since equality isn't defined on functions.  So rewrote those tests to test results instead of identity.

### DIFF
--- a/scalatest-test/src/test/scala/org/scalatest/prop/CommonGeneratorsSpec.scala
+++ b/scalatest-test/src/test/scala/org/scalatest/prop/CommonGeneratorsSpec.scala
@@ -3389,6 +3389,22 @@ If it doesn't show up for a while, please delete this comment.
       }
     }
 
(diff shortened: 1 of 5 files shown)


bvenners:
Closing this because cherry-picked all but one commit to this branch: https://github.com/scalatest/scalatest/tree/feature-implement-shrink-for-forall-with-one-arg

Will make a PR from that branch.

PR #1523, opened 2019-02-14
Shrink forAll arity 1; Fix forAll error message; cherry-pick Option/Either/Or PR

bvenners:


scalatest-test/src/test/scala/org/scalatest/enablers/PropCheckerAssertingAsyncSpec.scala
```diff
@@ -0,0 +1,117 @@
...
+import prop.GeneratorDrivenPropertyChecks
+import org.scalactic.anyvals.PosZInt
+import exceptions.TestFailedException
+import OptionValues._
+import scala.concurrent.Future
+
+class PropCheckerAssertingAsyncSpec extends AsyncFunSpec with Matchers with GeneratorDrivenPropertyChecks with LineNumberHelper {
+
+  describe("PropCheckerAsserting") {
+    it("The exception reported by an async forAll's Future should include the exception thrown by the property function evaluation as its cause") {
```
jducoeur: Note: we had a long conversation about this test pattern, and the fact that it isn't obvious why it works.  We should state explicitly, in the forAll() documentation, that if the function returns a Future, the next loop will not be tried until after that Future completes.  So the first loop to fail (even in a Future) will be the one returned by forAll().

jducoeur: Also worth noting that this means that forAll() over Futures is *never* going to be parallel (regardless of the EC used); don't expect that to help with your test performance.

scalatest/src/main/scala/org/scalatest/enablers/PropCheckerAsserting.scala
```diff
@@ -719,7 +726,34 @@ trait FuturePropCheckerAsserting {
...
 
-      case class AccumulatedResult(succeededCount: Int, discardedCount: Int, edges: List[A], rnd: Randomizer, initialSizes: List[PosZInt], result: Option[PropertyCheckResult])
+      case class AccumulatedResult(succeededCount: Int, discardedCount: Int, edges: List[A], rnd: Randomizer, initialSizes: List[PosZInt], result: Option[PropertyCheckResult], failedA: Option[A])
+
+      def shrunkenFuture(future: Future[PropertyCheckResult], a: A, rnd: Randomizer): Future[PropertyCheckResult] =
+        future.flatMap {
+          case pcr @ PropertyCheckResult.Failure(succeededCount, optEx, _, argsPassed, initSeed) => 
+            def shrinkLoop(shrinksRemaining: List[A]): Future[PropertyCheckResult] = {
+              shrinksRemaining match {
+                case Nil => Future.successful(pcr) // Can I reuse future here out of curiosity? That one is also completed.
```
jducoeur: To your comment: I believe that would result in the future flatMapping over *itself*.  No clue what would happen, but that seems kind of scary...

scalatest/src/main/scala/org/scalatest/enablers/PropCheckerAsserting.scala
```diff
@@ -1600,6 +1625,44 @@ object PropCheckerAsserting extends ExpectationPropCheckerAsserting with FutureP
...
+          case Some(sd: StackDepth) if sd.failedCodeFileNameAndLineNumberString.isDefined =>
+            "  " + FailureMessages.thrownExceptionsLocation(prettifier, UnquotedString(sd.failedCodeFileNameAndLineNumberString.get)) + EOL
+          case _ => ""
+        }
+      ) +
+      "  " + FailureMessages.occurredOnValues + EOL +
+      prettyArgs(getArgsWithSpecifiedNames(argNames, argsPassed), prettifier) + EOL +
+      "  )" +
+      getLabelDisplay(labels.toSet) + EOL +
+      "  " + FailureMessages.initSeed(prettifier, initSeed),
```
jducoeur: Just noticed - the trailing comma here has broken the build, and is a merge-blocker.  (It's legal in 2.12, but not before then.)  Please fix.

scalatest/src/main/scala/org/scalatest/enablers/PropCheckerAsserting.scala
```diff
@@ -719,7 +726,34 @@ trait FuturePropCheckerAsserting {
...
+
+      def shrunkenFuture(future: Future[PropertyCheckResult], a: A, rnd: Randomizer): Future[PropertyCheckResult] =
+        future.flatMap {
+          case pcr @ PropertyCheckResult.Failure(succeededCount, optEx, _, argsPassed, initSeed) => 
+            def shrinkLoop(shrinksRemaining: List[A]): Future[PropertyCheckResult] = {
+              shrinksRemaining match {
+                case Nil => Future.successful(pcr) // Can I reuse future here out of curiosity? That one is also completed.
+                case shrinkHead :: shrinkTail =>
+                  val result: Future[T] = fun(shrinkHead)
+                  result.transformWith {
```
jducoeur: This isn't legal -- `transformWith` is new in 2.12.

PR #1526, opened 2019-02-16
Added third-party Future support for asynchronous testing

junary:
We are using `com.twitter.util.Future`, and convert it to `scala.concurrent.Future`
when we are testing.

This pull request change `scala.concurrent.Future` dependencies of ScalaTest as
a drop-in replacement for asynchronous testing.

Here's an example:

```scala
    package org.scalatest.plus.example
   
    import com.twitter.util.{Future, Return, Throw}
    import org.scalatest.compatible.Assertion
    import org.scalatest.{AsyncFlatSpecLike, compatible}
   
    abstract class TwitterAsyncFlatSpec extends AsyncFlatSpecLike {
      // TODO: `Suite` is private[scalatest]
      override def toString: String = Suite.suiteToString(None, this)

      override type FutureSystem = Future[compatible.Assertion]
   
      override def convertToScalaFuture(f: FutureSystem): scala.concurrent.Future[compatible.Assertion] = {
        val promise: scala.concurrent.Promise[Assertion] = scala.concurrent.Promise()
        f.respond {
          case Return(value)    => promise.success(value)
          case Throw(exception) => promise.failure(exception)
        }
        promise.future
      }
    }
```


plinlor:
@cla-bot[bot] check

PR #1527, opened 2019-02-18
Dotty Build Step 1

cheeseng:
First batch of changes to support build for Dotty.


scalatest/src/main/scala/org/scalatest/Matchers.scala
```diff
@@ -3063,14 +3063,20 @@ trait Matchers extends Assertions with Tolerance with ShouldVerb with MatcherWor
...
 
-  private sealed trait Collected extends Product with Serializable
-  private case object AllCollected extends Collected
-  private case object EveryCollected extends Collected
-  private case class BetweenCollected(from: Int, to: Int) extends Collected
-  private case class AtLeastCollected(num: Int) extends Collected
-  private case class AtMostCollected(num: Int) extends Collected
-  private case object NoCollected extends Collected
-  private case class ExactlyCollected(num: Int) extends Collected
+  // In Dotty when Collected trait is private, object Matchers extends Matchers does not work, hopefully it can be fixed in future version of Dotty before it reaches final.
```
smarter: This is a deliberate change and not a bug: if `Collected` is private, then it should not appear in the signatures of methods and classes with a less restrictive access (like https://github.com/scalatest/scalatest/blob/fc3e0d259b5a9f1918f4f20c1b899c38e76c1681/scalatest/src/main/scala/org/scalatest/Matchers.scala#L6097 which is public).
Otherwise, the privateness of the type cannot be implemented properly: https://github.com/lampepfl/dotty/issues/4154.

liufengyun: If we agree with the approach in https://github.com/lampepfl/dotty/issues/4154, then we can close  https://github.com/lampepfl/dotty/issues/5831

cheeseng: Yes agree, making it to raise compiler error makes sense even under -language:Scala2 imho.

PR #1530, opened 2019-02-21
Matchers Collected Refactor

cheeseng:
This is a breaking change, the purpose of this refactoring is to fix the leak of the private Collected out through the constructor of the DSL classes in the Matchers trait.

Scala 2 compiler should have never allowed it from the beginning, in my humble opinion.  Dotty prevents this, so I think we should fix it in Scala 2 version also.

The breaking of the DSL classes compiler may be justified as: 

1) we mentioned in their comments that the classes are part of DSL and we do not really expect ScalaTest user to use it.
2) This is for 3.1.

### DIFF
--- a/project/GenScalaTestNative.scala
+++ b/project/GenScalaTestNative.scala
@@ -592,7 +592,11 @@ object GenScalaTestNative {
     //     "DeprecatedTimeLimitedTestsSpec.scala",   // skipped because DeprecatedTimeLimitedTests not supported.
     //     "TimeoutsSpec.scala"            // skipped because Timeouts not supported.
     //   )) ++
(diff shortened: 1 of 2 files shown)


jducoeur:
Agreed that I'm surprised this was ever legal, and this seems like the right time to back-port the fix to Scala 2.

bvenners:
@cheeseng we usually don't break anything except major releases, like 2.0, 3.0, 4.0. The only thing we do in minor releases, like 3.1, is remove things that have been deprecated. This has not been deprecated and would be a pain to do so.  Since it is already public, essentially, I just made the original Collected class public. That's non-breaking. Please take a look at this PR https://github.com/scalatest/scalatest/pull/1531  and try it with Dotty to make sure Dotty is happy with it. if so, and if you have no other concern, please go ahead and merge this into 3.1.x. Tnx.

PR #1531, opened 2019-02-21
Cheeseng feature matchers collected refactor

bvenners:
Make Collected public, since it was being returned from a public field in several case classes.

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/Matchers.scala
+++ b/scalatest/src/main/scala/org/scalatest/Matchers.scala
@@ -3063,7 +3063,7 @@ trait Matchers extends Assertions with Tolerance with ShouldVerb with MatcherWor
 
   // This is where InspectorShorthands started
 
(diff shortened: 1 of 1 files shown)


cheeseng:
@bvenners Oh, I didn't think of that direction, that way could be better I think, since Collected is public now, would you consider to make its sealed sub classes to public/protected (actually is protected better for Collected?)?  Which may enable user of Matchers to enhance the DSL themselves.

Anyway, I'll test this out with Dotty and it works, shall merge it first.

Thanks!

cheeseng:
@bvenners @jducoeur 

Tested: Both public and protected works on Dotty, what do you think of the idea of making it protected instead of public?

bvenners:
@cheeseng i didn't think of protected. That would be best probably. Go ahead and make that change then assign it back to me. Tnx.

cheeseng:
@bvenners I submitted a PR against your branch here: 

https://github.com/scalatest/scalatest/pull/1532

Thanks!

PR #1534, opened 2019-02-25
Improve startup performance by O(100)ms

eatkins:
Scala reflection is very slow because it takes a long time to load all
of the classes needed by the runtime mirror. Java reflection, on the
other hand, is reasonably quick. In unscientific and ad-hoc benchmarkes,
I found that the startup time decreased by about 100ms with this change.
Given java reflection was already used for scala 2.10, the risk of this
change seems minimal.


eatkins:
The test failures seem unrelated to this change.

exoego:
This sounds great !.

@eatkins 
Perhaps merging latest branch into this branch resolve test failures.

eatkins:
Actually, this seems to have been fixed in 15af09c5838b674ecb7e896569b4e42623a58899.

Issue #1524, opened 2019-02-14
Fill in missing canonicals and shrink methods


bvenners:
We are missing canonicals and shrink methods in the Generators for many types. I filled these in for NonZeroInt here for my Scala SF demos:

https://github.com/scalatest/scalatest/pull/1521/files#diff-073672db12d30d3b6a2c2ce74d3ccb37R1312

For example, the generator for nonZeroLong is missing these methods:

https://github.com/scalatest/scalatest/blob/2a9f13f163ec6766a8d26c6c3d554580f871ca2f/scalatest/src/main/scala/org/scalatest/prop/Generator.scala#L1328

Please ensure these exist for all the Generators we are providing out of the box. Please make tests for each. For example, the ones I wrote for NonZeroInt is here:

https://github.com/scalatest/scalatest/blob/2a9f13f163ec6766a8d26c6c3d554580f871ca2f/scalatest-test/src/test/scala/org/scalatest/prop/GeneratorSpec.scala#L1725

and here:

https://github.com/scalatest/scalatest/blob/2a9f13f163ec6766a8d26c6c3d554580f871ca2f/scalatest-test/src/test/scala/org/scalatest/prop/GeneratorSpec.scala#L1731

Thanks.

Bill

PR #1537, opened 2019-02-26
#1524: fill in missing shrink and canonicals methods

jducoeur:
This one's pretty big, but is mostly all the same thing: going through most of the Generators and adding `canonicals()` and `shrink()`.  It's a huge amount of code, nearly all of it in `Generator` and `GeneratorSpec`.

Broadly speaking, I've tried to keep to the same patterns as the existing implementations.  In particular, the numeric implementations all look pretty similar, and the collection implementations all look pretty similar.  To enforce this, I refactored a couple of methods out on the test side.

With 20/20 hindsight, I don't actually love the way this came out: there is *way* too much duplicate code in the `shrink()` implementations.  I think that there is a common typeclass desperately fighting to emerge for the Scalactic anyval types -- coming up with a single common implementation of `shrink()` would be a great way to guide what that might look like.  But that didn't occur to me until I was halfway through, and realized just how many of these types there really are.


bvenners:
This is probably the way it worked before, but I noticed our "strategery" algo is doing this:

```
scala> it.toList
res11: List[List[Int]] = List(List(), List(0), List(1), List(-1), List(2), List(-2), List(3), List(-3), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100), List(100, 100), List(100, 100, 100, 100), List(100, 100, 100, 100, 100, 100, 100, 100), List(100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100, 100))
```

It repeats List(100) a bunch of times.

jducoeur:
Hmm.  Annoying, and underlines why we need to get better about factoring this stuff properly.  I tried to factor the test code to reduce duplication, but didn't have time to think about how to refactor the collection `shrink()` implementations.  I'd recommend figuring out this bug, but instead of fixing it individually for each type (since I expect the same issue to show up for each collection type), instead refactoring those methods and fixing it once...

PR #1543, opened 2019-03-07
Adjustments for 3.0.7

cheeseng:
Fixed binary compatibility problem from 3.0.5 -> 3.0.6, getting ready to be released as 3.0.7.


scalatest/src/main/scala/org/scalatestplus/package.scala
```diff
@@ -0,0 +1,146 @@
...
+}
+
+/**
+ * Package objects of type alias for deprecated mockito classes, there will be moved in these new packages in the future version of ScalaTest.
+ */
+package object mockito {
+
+  type MockitoSugar = org.scalatest.mockito.MockitoSugar
+
+  lazy val MockitoSugar = org.scalatest.mockito.MockitoSugar
```
bvenners: @cheeseng Any reason these are lazy vals instead of just plain old vals?

cheeseng: @cheeseng Not special reason, just get used to lazy val already for non-local values. :)

PR #1559, opened 2019-03-26
Dotty Adjustments

cheeseng:
@liufengyun @bvenners 

Adjustments for latest Dotty.

Note that while AccumulationSpec now works, we have a regression in DefaultEqualitySpec (error during macro expansion).

Good news is the worrying compiler hanging problem now is fixed, though there're compiler error that's required further investigation.

### DIFF
--- a/scalatest-test/src/test/scala/org/scalatest/AssertionsSpec.scala
+++ b/scalatest-test/src/test/scala/org/scalatest/AssertionsSpec.scala
@@ -71,7 +71,6 @@ class AssertionsSpec extends AnyFunSpec {
         assert(a3 === a1)
       }
     }
-    // SKIP-DOTTY-START
     it("should compare nulls in a satisfying manner") {
       val n1: String = null
       val n2: String = null
@@ -95,7 +94,6 @@ class AssertionsSpec extends AnyFunSpec {
       }
       assert(e1.message === Some(FailureMessages.didNotEqual(prettifier, a, null)))
     }
-    // SKIP-DOTTY-END
   }
   describe("The intercept method") {
     it("should catch subtypes") {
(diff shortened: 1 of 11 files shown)


liufengyun:
The hang is fixed by https://github.com/lampepfl/dotty/pull/6163 and confirmed by the minimized test  https://github.com/lampepfl/dotty/pull/6166 .

The compilation error is due to this one: https://github.com/lampepfl/dotty/issues/6114

liufengyun:
FYI: the regression in `DefaultEqualitySepc` is fixed in https://github.com/lampepfl/dotty/pull/6176, should compile tomorrow.

cheeseng:
@liufengyun Cool!  I shall test it when it becomes available tomorrow, and remove the DefaultEqualitySpec in this PR once it is working.

Thanks for the info and update!

scalatest-test/src/test/scala/org/scalatest/AssertionsSpec.scala
```diff
@@ -4969,25 +4971,25 @@ class AssertionsSpec extends AnyFunSpec {
...
       val e = intercept[TestCanceledException] {
-        assume(null == a, ". dude")
+        assume(null == s, ". dude")
       }
-      assert(e.message === Some(didNotEqual(null, 3) + ". dude"))
+      assert(e.message === Some(didNotEqual(null, "test") + ". dude"))
       assert(e.failedCodeFileName === (Some(fileName)))
       assert(e.failedCodeLineNumber === (Some(thisLineNumber - 4)))
     }
-    // SKIP-DOTTY-END
```
liufengyun: Nice to remove the dotty skips 👍

cheeseng:
@liufengyun fyi I made some progress on your CompileMacro implementation: 

https://github.com/scalatest/scalatest/pull/1559/commits/f1a6cbfc13a1c2e97bd174e7810d95976be1409d

cheeseng:
@liufengyun A quick question, I wonder is there a way to differentiate a parse and type error?

Thanks!

liufengyun:
@cheeseng Currently we don't have that, we are assuming library authors only interested in type errors.

Are there use cases for parsing errors? That seems to be testing the compiler instead of library.

cheeseng:
@liufengyun Currently ScalaTest just give a different error message when it is failing, e.g. it is a parse/type error.  I guess it is fine if it is too much trouble, just a nice to have thing.  :)

PR #1565, opened 2019-04-03
[3.1.x] Scalactic Without Scala-xml

cheeseng:
Removed scala-xml dependency from scalactic.


cheeseng:
Closing, as Prettifier still needs xml support.

PR #1566, opened 2019-04-03
[3.0.x] Scalactic Without Scala-xml 

cheeseng:
Removed scala-xml dependency from scalactic.

### DIFF
--- a/project/scalatest.scala
+++ b/project/scalatest.scala
@@ -365,7 +365,6 @@ object ScalatestBuild extends Build {
       projectTitle := "Scalactic",
       organization := "org.scalactic",
       initialCommands in console := "import org.scalactic._",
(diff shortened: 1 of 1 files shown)


cheeseng:
Actually Prettifier is using xml classes, not sure why it builds in 2.12 and travis, will investigate further.

cheeseng:
Ok, found the reason, scala-compiler 2.12 depends on scala-xml: 

https://mvnrepository.com/artifact/org.scala-lang/scala-compiler/2.12.8

while 2.13 does not: 

https://mvnrepository.com/artifact/org.scala-lang/scala-compiler/2.13.0-M5

cheeseng:
Closing, as Prettifier still needs xml support.

PR #1572, opened 2019-04-08
Dotty - pass expectation specs

liufengyun:
only 6 tests skipped, which is not supported in Dotty

- 3 about checking parsing errors
- 3 about checking type error messages

### DIFF
--- a/scalatest.dotty/src/main/scala/org/scalatest/Expectations.scala
+++ b/scalatest.dotty/src/main/scala/org/scalatest/Expectations.scala
@@ -21,7 +21,13 @@ import scala.concurrent.ExecutionContext
 import scala.concurrent.Future
 import scala.reflect.ClassTag
 
+import scala.tasty._
+import scala.quoted._
+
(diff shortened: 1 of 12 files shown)


scalatest.dotty/src/main/scala/org/scalatest/Expectations.scala
```diff
@@ -21,8 +21,14 @@ import scala.concurrent.ExecutionContext
...
 import scala.reflect.ClassTag
 
+import scala.tasty._
+import scala.quoted._
+
 private[scalatest] trait Expectations {
 
+  inline def (inline x: String) stripMargin <: String =
+    ${ Expectations.stripMarginImpl(x) }
+
```
liufengyun: @cheeseng I'm not sure if this is the right place to put the extension method.

This compile-time `stripMargin` can co-exist with runtime `stripMargin` without problem, as the folllowing PR shows:

https://github.com/lampepfl/dotty/pull/6262

PR #1578, opened 2019-04-17
change firefox driver constructor for compatibility with newer versions

NikolajsCmihuns:
Hi, discovered that FirefoxDriver contructor
`public FirefoxDriver(FirefoxProfile profile)`
with used in 3.0.x branches is from selenium-firefox-driver:2.45.0 version (which is too old) and no longer exists in newer versions. In modern projects version of selenium-firefox-driver will be higher and when initialisation of [FirefoxDriver](https://github.com/scalatest/scalatest/blob/c35b47a8b32d6f6522690523b7d48d95666b664d/scalatest/src/main/scala/org/scalatestplus/selenium/WebBrowser.scala#L4523) happens, then NoSuchMethodError is thrown.

@bvenners  @cheeseng Maybe we can update selenium versions in scalatest project itself ?

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/selenium/WebBrowser.scala
+++ b/scalatest/src/main/scala/org/scalatest/selenium/WebBrowser.scala
@@ -4509,25 +4509,9 @@ object HtmlUnit extends HtmlUnit
 trait Firefox extends WebBrowser with Driver with ScreenshotCapturer {
 
   /**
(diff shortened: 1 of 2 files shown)


plinlor:
@cla-bot[bot] check

artimasites:
@cla-bot[bot] check

PR #1581, opened 2019-04-27
Make Equivalence serialisable

bbonanno:
This helps when you capture an instance of Equality/Equivalence in some class that has to be serialised.

In my particular case, I use it to match invocation arguments on mocks in mockito-scala and for some scenarios said mocks have to be serialisable.


plinlor:
@cla-bot[bot] check

PR #1588, opened 2019-05-19
[3.1.x] Scalactic Scala-XML Removal

cheeseng:
Backported from 3.0.x, removed scala-xml dependency from Prettifier, thus removing it from scalactic also.


smarter:
Do you plan to merge this soon? I'd like to remove scala-xml from the dependencies of Dotty but this will break the scalactic.dotty until this is merged. /cc @liufengyun

cheeseng:
@smarter Fyi I have proposed a PR list to @bvenners for merging, and this is one of them (and the one from fengyun too), hopefully @bvenners will be able to merge this in soon.

Cheers.

PR #1590, opened 2019-05-22
scalactic.dotty: add scala-xml dependency

smarter:
This wasn't needed so far because dotty-compiler depends on scala-xml
but that will change soon.

/cc @liufengyun


smarter:
Oh, I guess this won't be needed once https://github.com/scalatest/scalatest/pull/1588 is merged so closing.

PR #1591, opened 2019-05-24
Fix stray timer in SuiteSortingReporter

### LINKED ISSUE #1175: java.util.TimerThread not stopped after test is finished
After running multiple times the same test inside sbt on my project, I got a Metaspace OOM error. I removed every pending thread and at the end only the "Timer-0" thread survived at the end of the test. 

To make the investigation easier, I created a very small project with just two tests 
[https://github.com/giampaolotrapasso/TimerIssue](https://github.com/giampaolotrapasso/TimerIssue) and I repeated the `test` command five times.

As you can see from the picture, every time a new timer get allocated without removing the previous one.
threads
HProf files are [available](https://github.com/giampaolotrapasso/TimerIssue/tree/master/hprof)

My very rough analysis -not scalatest expert at all- concluded that this pending thread should cause the metaspace retention and so the OOM. The timer is fired inside `SuiteSortingReporter` that is allocated with the [sbt test task](https://github.com/scalatest/scalatest/blob/208278abc69cc4bcb03c6bd4e25b876211678520/scalatest/src/main/scala/org/scalatest/tools/Framework.scala#L666) 

Any thoughts on this? Did I misunderstood something?
Thank you in advance

gleber-da:
This should fix #1175

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/tools/SuiteSortingReporter.scala
+++ b/scalatest/src/main/scala/org/scalatest/tools/SuiteSortingReporter.scala
@@ -285,11 +285,13 @@ private[scalatest] class SuiteSortingReporter(dispatch: Reporter, val testSortin
     timeoutTask match {
         case Some(task) => 
           if (head.suiteId != task.slot.suiteId) {
+            timer.cancel()
             task.cancel()
             timeoutTask = Some(new TimeoutTask(head)) // Replace the old with the new
             timer.schedule(timeoutTask.get, testSortingTimeout.millisPart)
           }
         case None => 
+          timer.cancel()
           timeoutTask = Some(new TimeoutTask(head)) // Just create a new one
           timer.schedule(timeoutTask.get, testSortingTimeout.millisPart)
       }


gleber-da:
@cheeseng Please let me know if there's a better way to fix it.

cheeseng:
@gleber-da Sorry for attending to this late, I look into this but I am not sure if a timer.schedule call is allowed after a timer.cancel() is called, according to the schedule method javadoc here: 

https://docs.oracle.com/javase/8/docs/api/java/util/Timer.html#schedule-java.util.TimerTask-long-

It shall get a IllegalStateException ?  I didn't try it yet but I think it is good idea to believe what javadoc says.  :)

It seems to be doDispose is a more suitable place for calling timer.cancel(), would that work?

cheeseng:
@gleber-da Fyi I added an alternative PR here: 

https://github.com/scalatest/scalatest/pull/1774

It'll be great if you can take a look to see if it makes sense to you, or if you can help to verify if the fix works it would be even greater!

cheeseng:
Superseded by https://github.com/scalatest/scalatest/pull/1774

Thank you!

PR #1592, opened 2019-05-26
[3.0.x] SBT @WrapWith Fix

cheeseng:
Changed Framework to flavour @WrapWith when it is available, consistent with behaviour of ScalaTestFramework and Runner.

Note: This fix should be cherry-picked into 3.1.x if it is accepted.

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/tools/Framework.scala
+++ b/scalatest/src/main/scala/org/scalatest/tools/Framework.scala
@@ -434,9 +434,7 @@ class Framework extends SbtFramework {
       if (accessible || runnable) {
         val suite =
           try {
-            if (accessible)
-              suiteClass.newInstance.asInstanceOf[Suite]
-            else {
+            if (runnable) { // When it is runnable WrapWith is available, this will take precedence and this behavior will be consistent with Runner and the old ScalaTestFramework.
               val wrapWithAnnotation = suiteClass.getAnnotation(classOf[WrapWith])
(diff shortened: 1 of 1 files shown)


christian-schlichtherle:
Has this been tested with both SBT and the scalatest-maven-plugin?

christian-schlichtherle:
Also, this changes the behavior in one place. But when looked for usages of the `WrapWith` annotation in the source code, I found about six places which are copy/pasted. What about the other five code locations?

christian-schlichtherle:
Last, but not least: Thank you very much for your efforts to look into this!

cheeseng:
@christian-schlichtherle Thanks for the note, I am not sure about all the six places that you're referring to, but fyi in ScalaTest there are 3 main 'runner' are available, which are: 

org/scalatest/tools/Framework.scala - Runner for current version of SBT, more precisely SBT version that started to support test-interface (starting SBT 0.12/0.13, can't recall offhead at the moment)

org/scalatest/tools/ScalaTestFramework.scala - Runner for older version of SBT, pre test-interface versions.

org/scalatest/tools/Runner - Runner for most of the other things, which includes, maven, ant, intellij, eclipse etc., or from command line directly.

Other code that is dealing with @WrapWith probably is used by the above 3 runners in some way, Runner, ScalaTestFramework (the old) has the intended behavior, it is just the Framework (newer) that's is inconsistent and that's what being fixed by this PR.  :)

PR #1607, opened 2019-06-18
 Use Select.apply instead of Select.overload in Dotty macros

liufengyun:
Use Select.apply instead of Select.overload in Dotty macros

```Scala
- val res = TypeApply(Select.unique(Select.unique(l, "value"), op), targs).seal.cast[R]
+ val res = Select.unique(l, "value").select(sel.symbol).appliedToTypes(targs.map(_.tpe)).seal.cast[R]
```

Previously the code above assumes the member `op` is unique, which is potentially problematic. Now use symbolic select, it's more robust.

### DIFF
--- a/project/GenScalacticDotty.scala
+++ b/project/GenScalacticDotty.scala
@@ -160,7 +160,8 @@ object GenScalacticDotty {
         "PrettifierSpec.scala",  // Test failed with java.lang.IllegalAccessException
         "RequirementsSpec.scala",  // Error during macro expansion
         "TolerantEquivalenceSpec.scala",   // Compilation error to be investigated further.
-        "TripleEqualsSpec.for210"  // Old staff, we shall delete this soon.
+        "TripleEqualsSpec.for210",  // Old staff, we shall delete this soon.
+        "FutureSugarSpec.scala"     // instability, occasional timeout in CI
       )) ++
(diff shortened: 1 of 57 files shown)


liufengyun:
@bvenners @cheeseng With the last commit, we run the whole test set of scalactic, except a few due to

- bugs or changes in Dotty
- macros for `assert` needs to be improved for better messages for `.exists`, `.length`, `.size`, `.isInstanceOf`, `.isEmpty`.

liufengyun:
@bvenners @cheeseng Now the assertion macro is on par with Scala2 after 281e06e  and bbc0975.

cheeseng:
@liufengyun  Whao, that's great news!  I'll try it out as soon as possible.

cheeseng:
@liufengyun  After merging it seems to build fail under Scala 2.12, not sure why Travis didn't fail, I am working on a fix and shall submit a PR against your branch today.

Cheers.

liufengyun:
Thanks a lot @cheeseng !

cheeseng:
@liufengyun I submitted the following PR to your branch: 

https://github.com/liufengyun/scalatest/pull/6

and I figured out the reason Travis didn't fail previously: 

https://github.com/liufengyun/scalatest/pull/6/commits/0e2fc4d854dd91be63072c5262b7d3913a3e6539

Cheers.

PR #1609, opened 2019-06-19
Privatize Generator

bvenners:
@cheeseng One problem here is that I commented out the test classes for NumericString and RegexString, because they can't access Generator now that it is private[scalatest]. The other test classes are using implicit generators that are defined in the Generator companion, and they don't have any trouble accessing. I don't see why they can access. My theory is we should need to make Generator private[org]. Anyway, please take a look at that. We want to resurrect the RegexString and NumericString test classes.

### DIFF
--- a/examples/src/test/scala/org/scalatest/examples/propspec/multi/ExampleSpec.scala
+++ b/examples/src/test/scala/org/scalatest/examples/propspec/multi/ExampleSpec.scala
@@ -19,7 +19,7 @@ import org.scalatest._
 import prop.PropertyChecks
 import scala.collection.mutable.ListBuffer
 
-class ExampleSpec extends fixture.PropSpec with PropertyChecks with Matchers {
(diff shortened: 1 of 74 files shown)


cheeseng:
@bvenners I wonder if we move our generator things into common-test will work better, which comes with the advantage that our test classes can still access it and these classes will not be included in the built jar files.  What do you think?

cheeseng:
@bvenners I would like to suggest also to make a branch off for next version after 3.1, e.g. next-version-after-3.1, so we do not need to move the generator thing in again in the next release where we probably want to include our generator.

bvenners:
@cheeseng both of those ideas sound good: 1. make a branch to save the current state of 3.1.x, then 2. move our generator to common-test. I hadn't thought of the latter, but that's better.

cheeseng:
Closing this, we'll go with common-test approach.

Thanks.

PR #1610, opened 2019-06-21
Move ScalaTest generator to common-test for the 3.1.0 release

bvenners:


cheeseng:
Hi @bvenners , I tested this but it seems like scala-js build (and probably scala-native) is not working yet, I can help in fixing it when possible.

Cheers.

Issue #1605, opened 2019-06-10
Intermittent ConcurrentModificationException with AsyncFreeSpec


mpilquist:
We ran in to this issue in FS2 when using 3.1.0-SNAP11. A minimized version:

```scala
package scratch

import scala.concurrent._

import org.scalatest.Succeeded
import org.scalatest.freespec.AsyncFreeSpec

class Repo extends AsyncFreeSpec {

  implicit override val executionContext: ExecutionContext =
    scala.concurrent.ExecutionContext.Implicits.global

  ('a' to 'z').map { x =>
    x.toString - {
      (1 to 10).map(i => i.toString in Future(Succeeded))
    }
  }
}
```

Sometimes this test fails with:

```
java.util.ConcurrentModificationException: Two threads have apparently attempted to run a suite at the same time. This has resulted in both threads attempting to concurrently change the current Informer. Suite class name: scratch.Repo
```

We worked around this issue in FS2 by using the default sequential `ExecutionContext` for test registration and the normal global context for everything else. Here's our workaround, which seems to work: https://github.com/functional-streams-for-scala/fs2/blob/1ff53eb4988535f29bbedfa76fb8b423b6dba415/core/shared/src/test/scala/fs2/Fs2Spec.scala#L40-L46

PR #1619, opened 2019-07-03
Concurrency Problem Fix

cheeseng:
Rewritten StatefulStatus, ScalaTestStatefulStatus and AsyncOutcome without synchronized calls, this fixes the problem as reported here: 

https://github.com/scalatest/scalatest/issues/1605

Note: I think we shall backport this to 3.0.x.


cheeseng:
Superseded by another PR submitted by @bvenners , which is now merged into master.

PR #1625, opened 2019-07-09
Avoid escaped characters in index.html

daddykotex:
The reporter in scala 2.11.12, generates invalid html. This happens because the content of CDATA is escaped when it should not be.

Screen Shot 2019-07-09 at 14 37 48

This is a fix for it. I'm open to suggestion because I'm assuming this is an issue in scala-xml rather than this code. But in the meantime, that would be the fix.


daddykotex:
Anything might come out of this?

plinlor:
@cla-bot[bot] check

artimasites:
@cla-bot[bot] check

daddykotex:
Signed

PR #1643, opened 2019-07-19
Move Expectations to org.scalatest.expectations package

bvenners:


bvenners:
This still needs work in the build for Scala.js and native. Closing.

PR #1661, opened 2019-07-26
Green 2.13 Tests

cheeseng:
- Get tests to green when builds with Scala 2.13.0.
- Make Scala 2.13.0 the default scala version.
- Added Scala 2.13.0 to Travis Build.

PR #1663, opened 2019-07-27
Get tests green under 2.13

bvenners:
Replaces https://github.com/scalatest/scalatest/pull/1661. 

Main differences between this and PR1661:
- Tests for `contain.only()` under 2.13 also.
- Doesn't paper over breaking change around `Float` and `Double` `Ordering` in 2.13.
- Doesn't paper over Scala.js change around `InterruptedException` and `Future.apply`.
- Changes `initialCommands` in the build to use 3.1.0 name for `should.Matchers`.
- Modifies build so that sbt doc works in 2.13.0, which is now the build default Scala version.


SethTisue:
is there any 3.0.x branch where the tests are expected to pass on 2.13? asking because of the community build.

PR #1664, opened 2019-07-29
Update examples and improve Scaladoc

bvenners:


bvenners:
Closing this given it has the problem with valueOf in the prop package in common-test, which was identified by Chee Seng. Will submit a new PR later.

PR #1665, opened 2019-07-29
Moving valueOf Function

cheeseng:
Moved valueOf function in common-test's org.scalatest.prop package object to CommonGenerators.


bvenners:
I don't think it is worth doing major surgery to get a line or two package documentation for 3.1.0. I will just comment out the documentation for now. The valueOf methods were in the package object to make error messages involving functions more concise. Putting them in CommonGenerators would defeat that purpose.

PR #1668, opened 2019-07-30
Update Dotty macros

nicolasstucki:
These are the same patches I added to the Dotty community build [scalatest/dotty-community-build](https://github.com/dotty-staging/scalatest/commits/dotty-community-build). These changes correspond to the changes introduced in Dotty 0.17.0-RC1.


cheeseng:
@nicolasstucki Unfortunately our latest refactoring of the package is breaking the build, I'll be working on fixing it and submit a PR against your branch.

PR #1681, opened 2019-08-16
JS Modules

cheeseng:
Generated modules for JS world.


bvenners:
Going to merge this, because the failures are unrelated to this PR I think. But they look like real problems that we should fix (problems in our generated tests, not in ScalaTest or Scalactic itself). This is the Travis build:

https://travis-ci.org/scalatest/scalatest/builds/572807623?utm_source=github_status&utm_medium=notification

Issue #1418, opened 2018-10-09
The here macro to materialize a source file position slows down compilation


retronym:
I'm [working](https://github.com/retronym/scala/pull/28) on instrumentation for the compiler to attribute compilation time to compilation of particular file/line/method/implicits/macros.

I've used this tool to analyse the compilation of https://github.com/guardian/frontend, which is a relatively large application that uses Play and ScalaTest.

One thing that stood out is that 1.5% of compilation time is attributed to the expansion of the `here` macro in ScalaTest, which is implemented by `genPosition`.

https://github.com/scalatest/scalatest/blob/a7673db180bf14e757d476fb06df4aa17948e1a8/scalactic-macro/src/main/scala/org/scalactic/source/PositionMacro.scala#L35-L63

It might be possible to reduce the cost of this macro, by reducing the number of calls to `newTermName` and avoiding generating a big nested select that must be typechecked to find the `Position` class.

Here's a more direct way to generate a call to `scala.Symbol.apply`, for instance.
```
scala> def impl(c: Context): c.Tree = { import c.universe._; Apply(typeOf[Symbol.type].decl(TermName("apply")), Literal(Constant("blah")))}; def m: Any = macro impl
<console>:28: warning: method Apply in trait Trees is deprecated (since 2.10.1): use q"$sym(..$args)" instead
       def impl(c: Context): c.Tree = { import c.universe._; Apply(typeOf[Symbol.type].decl(TermName("apply")), Literal(Constant("blah")))}; def m: Any = macro impl
                                                             ^
impl: (c: scala.reflect.macros.blackbox.Context)c.Tree
defined term macro m: Any

scala> m
res1: Any = 'blah
```

retronym:
![image](https://user-images.githubusercontent.com/65551/46644431-91cd1d00-cbc3-11e8-82b6-f34db17365a8.png)

bvenners:
@retronym Thanks for the idea (and the pretty diagram). I've assigned this to Chee Seng.

retronym:
I'm looking more closely for the bottleneck. So far, I noticed this stack trace:

```
"ForkJoinPool-11-worker-1" #25 daemon prio=5 os_prio=31 tid=0x00007fc9783fb000 nid=0xa617 runnable [0x000070000597a000]
   java.lang.Thread.State: RUNNABLE
	at java.lang.Throwable.fillInStackTrace(Native Method)
[...]

PR #1690, opened 2019-08-29
PositionMacro from Jason Zaugg

cheeseng:
@bvenners This is PositionMacro implementation by Jason Zaugg, as discussed here: 

https://github.com/scalatest/scalatest/issues/1418

I tried to measure the improvements of compile time for our tests and here's the result: 

## Scala 2.13.0

| Run | New with BlackBox | New without BlackBox | Current Version |
|-------|---------------------------|--------------------------------|----------------------|
| 1     | 266s                        | 260s                              |                 274s |
| 2     | 261s                        | 261s                              |                 274s |
| 3     | 265s                        | 263s                              |                 274s |
| Avg | 264s                        | 261.33s                         |                 274s |

## Scala 2.12.9

| Run | New with BlackBox | New without BlackBox | Current Version |
|-------|---------------------------|--------------------------------|----------------------|
| 1     | 256s                        | 246s                              |                 265s |
| 2     | 250s                        | 247s                              |                 265s |
| 3     | 253s                        | 249s                              |                 261s |
| Avg | 253s                        | 247.33s                         |            263.67s |

## Scala 2.12.8

| Run | New with BlackBox | New without BlackBox | Current Version |
|-------|---------------------------|--------------------------------|----------------------|
| 1     | 273s                        | 276s                              |                 287s |
| 2     | 277s                        | 273s                              |                 283s |
| 3     | 277s                        | 277s                              |                 283s |
| Avg | 275.67s                   | 275.33s                         |            284.33s |

In conclusion, there's definitely improvement for compile time for the new version (with and without BlackBox).  In https://github.com/scalatest/scalatest/issues/1418#issuecomment-507934810 , @retronym actually recommended to use BlackBox, which only available in Scala 2.12 and 2.13 only.  So in order to use BlackBox we shall generate PositionMacro based on Scala version, however, looking at the measurement above, in my humble opinion we may just use the version without BlackBox (as in this PR, which are not generating PositionMacro based on Scala version).

Anyway, do let me know if you would like the BlackBox version, I can generate PositionMacro in a new commit.

Thanks.


retronym:
Aside from a potential compile time performance change, there is another problem with the unnecessary usage of a whitebox-macro version of `PositionMacro`. It's a problem that I'm facing with a customer of ours using ScalaTest in a large build. Their build is intermittently failing with "implicit not found: Position" errors, but works on a subsequent build. One hypothesis is that the macro expansion is failing intermittently, perhaps due to some quirk in Scalac's macro classloader, or something else altogether. But problems expanding implicit whitebox macros are treated as "implicit not found" rather than "fatal error to report to the user", so I haven't been able to troubleshoot the root cause.

## Blackbox intermittent implicit macro error
```
scala> import reflect.macros._, language.experimental.macros
import reflect.macros._
import language.experimental.macros

scala> class C1; implicit def c1: C1 = macro c1Impl; def c1Impl(c: blackbox.Context) = {import c.universe._; if (c.macroApplication.hashCode % 4 == 0) throw null else q"null"}
defined class C1
defined term macro c1: C1
c1Impl: (c: scala.reflect.macros.blackbox.Context)c.universe.Literal

scala> implicitly[C1]
res0: C1 = null

scala> implicitly[C1]
res1: C1 = null

scala> implicitly[C1]
res2: C1 = null

scala> implicitly[C1]
res3: C1 = null

scala> implicitly[C1]
<console>:18: error: exception during macro expansion:
java.lang.NullPointerException
	at .c1Impl(<console>:15)

       implicitly[C1]
                 ^
```

## Whitebox intermittent implicit macro error

```
scala> import reflect.macros._, language.experimental.macros
import reflect.macros._
import language.experimental.macros

scala> class C1; implicit def c1: C1 = macro c1Impl; def c1Impl(c: whitebox.Context) = {import c.universe._; if (c.macroApplication.hashCode % 4 == 0) throw null else q"null"}
defined class C1
defined term macro c1: C1
c1Impl: (c: scala.reflect.macros.whitebox.Context)c.universe.Literal

scala> implicitly[C1]
res0: C1 = null

scala> implicitly[C1]
res1: C1 = null

scala> implicitly[C1]
res2: C1 = null

scala> implicitly[C1]
res3: C1 = null

scala> implicitly[C1]
res4: C1 = null

scala> implicitly[C1]
res5: C1 = null

scala> implicitly[C1]
<console>:18: error: could not find implicit value for parameter e: C1
       implicitly[C1]
                 ^
```

So while I understand it would add some complexity to your build, I still think using a blackbox macro is the right thing to do according the principle of using the least powerful tool to achieve the goal.

Perhaps you could structure your build with a single file that is Scala-version dependent that is defined as:

## Scala 2.11.x
```
object MacroContext {
  type BlackboxContext = scala.macros.Context
}
```

## Scala 2.12.x+
```
object MacroContext {
  type BlackboxContext = scala.macros.blackbox.Context
}
```

And change your the position macro implementation to use that type alias.

scalactic-macro/src/main/scala/org/scalactic/source/PositionMacro.scala
```diff
@@ -1,65 +1,75 @@
...
+  private class PositionMacroImpl(val universe: scala.reflect.api.Universe) {
+
+    private val PositionModule = universe.rootMirror.staticModule("org.scalactic.source.Position")
+    private val PositionClass = universe.rootMirror.staticClass("org.scalactic.source.Position")
+    private val Position_apply = PositionModule.typeSignature.declaration(universe.newTermName("apply"))
+
+    private[scalactic] lazy val showScalacticFillFilePathnames: Boolean = {
+      val value = System.getenv("SCALACTIC_FILL_FILE_PATHNAMES")
+      value != null && value == "yes"
+    }
```
retronym: `showScalacticFillFilePathnames` could still be a member of `PositionMacro`.

scalactic-macro/src/main/scala/org/scalactic/source/PositionMacro.scala
```diff
@@ -1,65 +1,75 @@
...
-  private[scalactic] lazy val showScalacticFillFilePathnames: Boolean = {
-    val value = System.getenv("SCALACTIC_FILL_FILE_PATHNAMES")
-    value != null && value == "yes"
+      // Our types don't capture the fact that context.universe eq this.universe, so we have to cast.
+      implicit def castSymbol(s: universe.Symbol): context.universe.Symbol = s.asInstanceOf[context.universe.Symbol]
+      implicit def castType(t: universe.Type): context.universe.Type = t.asInstanceOf[context.universe.Type]
+
+      import treeBuild.{mkAttributedIdent, mkAttributedSelect}
+
+      // Because this macro is called so frequently, we use assign types to the trees as we build them, rather
```
retronym: s/use assign/assign/

scalactic-macro/src/main/scala/org/scalactic/source/PositionMacro.scala
```diff
@@ -1,65 +1,75 @@
...
+      def strLit(s: String) = {
+        Literal(Constant(s)).setType(definitions.StringClass.toTypeConstructor)
+      }
+      def intLit(i: Int) = {
+        Literal(Constant(i)).setType(definitions.IntTpe)
+      }
+      val args = List(
+        strLit(context.enclosingPosition.source.file.name),
+        if (showScalacticFillFilePathnames) strLit(context.enclosingPosition.source.path) else strLit(Resources.pleaseDefineScalacticFillFilePathnameEnvVar), 
+        //strLit(""),
```
retronym: Remove commented code?

mkeskells:
I would certainly like to see this as a Blackbox macro

I am not sure that the benchmarks that you performed are representative in the real world
A whitebox macro will be inspected for every occasion that an implicit is possible
Making this BlackBox will reduce the complexity of this search, but you only see a benefit if the result was rejected  because the result type could never match

My contention would be that your tests would generally use the result of the macro, so there is normally no wasted time or resource

In real user code, I may be looking for an implicit Int (or any other type for that matter), and the scala compiler will generally attempt to expand this macro so see if it generates an Int, and when it doesn't, that resource (CPU and memory) is wasted. 
On the other hand, making this a Blackbox, means that the expansion will not occur as the result type could never match

@retronym can you confirm the above, or clarify/correct

If needed I m sure that we can generate a test case to demonstrate the cases where there would be significantly reduced resource by using Blackbox

retronym:
Whitebox macros are only expanded within implicit search if the declared type fits the expected type. So the white/black "boxity" of this implicit macro shouldn't affect the performance of unrelated implicit search at all.

Potentially, it could be wasteful when other an implicit search is weighing up candidate implicits that themselves accept an implicit parameter of type "Position".

Example:


```

cheeseng:
@retronym @mkeskells Thanks for the great detailed explanations!  Much appreciated and happy to learn!  We have only our tests (though quite a big set of them) to tests and benchmarked with, I simply didn't see the benefits before your posts, I'll work on making 2.12 and 2.13 to use blackbox next.

scalactic-macro/src/main/scala/org/scalactic/source/PositionMacro.scala
cheeseng: @retronym Sorry I don't get this, do you mind to clarify?

scalactic-macro/src/main/scala/org/scalactic/source/PositionMacro.scala
cheeseng: This is a good idea, thanks!

scalactic-macro/src/main/scala/org/scalactic/source/PositionMacro.scala
retronym: Change the comment to " // Because this macro is called so frequently, we use assign types as we build them, rather"

cheeseng: Updated comment: 

https://github.com/scalatest/scalatest/pull/1690/commits/5b338190018c665d7db3c1f185af7e0bcc818b15

PR #1711, opened 2019-10-05
Fix documentation of RightValue#value

markus1189:
Noticed that the docs for the right projection were wrong :)


scalatest/src/main/scala/org/scalatest/EitherValues.scala
```diff
@@ -141,7 +141,7 @@ trait EitherValues {
 
     /**
      * Returns the <code>Right</code> value contained in the wrapped <code>RightProjection</code>, if defined as a <code>Right</code>, else throws <code>TestFailedException</code> with
-     * a detail message indicating the <code>Either</code> was defined as a <code>Right</code>, not a <code>Left</code>.
+     * a detail message indicating the <code>Either</code> was defined as a <code>Left</code>, not a <code>Right</code>.
```
rtyley: This fix is also in #1775 - while doing that PR I also noticed it!

markus1189:
Seems like nobody cares :man_shrugging:

PR #1712, opened 2019-10-05
Add EitherValuable

markus1189:
In Scala 2.13.x, the `RightProjection` is deprecated (this time for good: https://github.com/scalatest/scalatest/pull/975).

So you can no longer write (without a deprecation warning about `right`):

```scala
val either: Either[String, Int] = Right(5)

either.value.right should ===(5)
```

So I added `EitherValuable` which has `rightValue` and `leftValue`.

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/EitherValues.scala
+++ b/scalatest/src/main/scala/org/scalatest/EitherValues.scala
@@ -112,13 +112,25 @@ trait EitherValues {
       * Returns the <code>Left</code> value contained in the wrapped <code>LeftProjection</code>, if defined as a <code>Left</code>, else throws <code>TestFailedException</code> with
       * a detail message indicating the <code>Either</code> was defined as a <code>Right</code>, not a <code>Left</code>.
       */
-    def leftValue: L = either.swap.getOrElse(throw new TestFailedException((_: StackDepthException) => Some(Resources.eitherLeftValueNotDefined), Some(new NoSuchElementException), pos))
+    def leftValue: L = {
+      try {
+        either.left.get
+      }
+      catch {
+        case cause: NoSuchElementException =>
+          throw new TestFailedException((_: StackDepthException) => Some(Resources.eitherLeftValueNotDefined), Some(cause), pos)
+      }
+    }
 
     /**
       * Returns the <code>Right</code> value contained in the wrapped <code>Either</code>, if defined as a <code>Right</code>, else throws <code>TestFailedException</code> with
       * a detail message indicating the <code>Either</code> was defined as a <code>Left</code>, not a <code>Right</code>.
       */
-    def rightValue: R = either.getOrElse(throw new TestFailedException((_: StackDepthException) => Some(Resources.eitherRightValueNotDefined), Some(new NoSuchElementException), pos))
+    def rightValue: R = either match {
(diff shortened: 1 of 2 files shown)


scalatest/src/main/scala/org/scalatest/EitherValues.scala
```diff
@@ -98,6 +107,32 @@ trait EitherValues {
...
   implicit def convertRightProjectionToValuable[L, R](rightProj: Either.RightProjection[L, R])(implicit pos: source.Position): RightValuable[L, R] = new RightValuable(rightProj, pos)
 
+  class EitherValuable[L, R](either: Either[L, R], pos: source.Position) {
+    /**
+      * Returns the <code>Left</code> value contained in the wrapped <code>LeftProjection</code>, if defined as a <code>Left</code>, else throws <code>TestFailedException</code> with
+      * a detail message indicating the <code>Either</code> was defined as a <code>Right</code>, not a <code>Left</code>.
+      */
+    def leftValue: L = {
+      try {
+        either.left.get
```
fancywriter: well, you are basically doing the same thing, which is deprecated and might disappear in newer Scala versions (2.14? 3.0?).
I though the intent was to do
```
either.swap.getOrElse(throw new TestFailedException(...))
```
without `try` block.

scalatest/src/main/scala/org/scalatest/EitherValues.scala
```diff
@@ -98,6 +107,32 @@ trait EitherValues {
...
+          throw new TestFailedException((_: StackDepthException) => Some(Resources.eitherLeftValueNotDefined), Some(cause), pos)
+      }
+    }
+
+    /**
+      * Returns the <code>Right</code> value contained in the wrapped <code>Either</code>, if defined as a <code>Right</code>, else throws <code>TestFailedException</code> with
+      * a detail message indicating the <code>Either</code> was defined as a <code>Left</code>, not a <code>Right</code>.
+      */
+    def rightValue: R = either match {
+      case Right(x) => x
```
fancywriter: and the same here, just `either.getOrElse(throw new TestFailedException(...))` should work.

scalatest/src/main/scala/org/scalatest/EitherValues.scala
markus1189: Only `.right.get` is deprecated because `Either` is now right biased by default.  The `.left` projections have been deprecated shortly but un-deprecated in https://github.com/scala/scala/pull/8012 :)

scalatest/src/main/scala/org/scalatest/EitherValues.scala
markus1189: `.getOrElse` is nicer than the pattern matching, fixed! :)

scalatest/src/main/scala/org/scalatest/EitherValues.scala
markus1189: But the `swap` followed by `getOrElse` is a bit nicer, so I used that now instead, thanks for the suggestion!

rtyley:
Oh, I opened #1775 without seeing this PR first, I think either approach is good.

markus1189:
I lost interest in this

Issue #1650, opened 2019-07-23
Report value of other side if a test using EitherValues is not the expected side


Mahoney:
At the moment if you mix in `EitherValues` and it is the wrong side:
```scala
test("either") {
  val either: Either[String, Int] = Left("Failed because 1 is not even")
  either.right.value mustBe 1
}
```
the message is:
`The Either on which right.value was invoked was not defined as a Right.`

It would be easier to understand a test failure if it were:
`The Either on which right.value was invoked was a Left with value "Failed because 1 is not even"`

Mahoney:
Raised https://github.com/scalatest/scalatest/pull/1738

jroper:
Agree, the only reason I wanted to use `EitherValues` was so that I didn't need to write the boilerplate to display what was on the left when treating it as a right, without this `EitherValues` seems to be next to useless.

bvenners:
merged. Tnx for the PR!

PR #1738, opened 2019-12-16
Report actual value of unexpected Either & Try projections

Mahoney:
More often than not when a test fails with "The Either on which
[left/right].value was invoked was not defined as a [Left/Right]"
or "The Try on which [failure/success] was invoked was not
defined as a [Failure/Success]" I immediately want to know what
it actually was - so why not report it?


Mahoney:
See https://github.com/scalatest/scalatest/issues/1650

cheeseng:
This looks like a nice enhancement to me.

or for Corporate:

Cheers.

Mahoney:
CLA signed & emailer, @cheeseng

PR #1749, opened 2020-01-03
Noinline higher order functions to speed up compilation in Scala.js

### LINKED ISSUE #1161: Add @noinline for faster Scala.js incremental linking times
It appears that the Scala.js linker tries to inline all `test("") {}` and `in {}` closures from ScalaTest, resulting in slow incremental linking times for large test suites. See related discussion:
https://github.com/scala-js/scala-js/issues/2827#issuecomment-302388572 In summary, it seems that adding `@noinline` to all methods that accept a `testFun: => Any` argument (example diff: https://github.com/dotty-staging/scalatest/commit/2a08a06fac79c991efbb6c59bb8d87466f0f82ed) can speed up incremental linking times for small edits. One informal benchmark from the scalameta/scalameta test suite (~1.000 unit tests) shows a 7x speedup, or 600ms instead of 4.5s.

Would you be open to accepting a PR adding `@noinline` to common methods like `def test{}` and `def in {}`?

exoego:
Closes #1161

I have not measured performance since I do not have testbed for slow compilation on Scala.js.

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/Assertions.scala
+++ b/scalatest/src/main/scala/org/scalatest/Assertions.scala
@@ -743,7 +743,7 @@ trait Assertions extends TripleEquals  {
    * @throws TestFailedException if the passed function does not complete abruptly with an exception
    *    that's an instance of the specified type.
    */
-  def intercept[T <: AnyRef](f: => Any)(implicit classTag: ClassTag[T], pos: source.Position): T = {
+  @noinline def intercept[T <: AnyRef](f: => Any)(implicit classTag: ClassTag[T], pos: source.Position): T = {
     val clazz = classTag.runtimeClass
     val caught = try {
       f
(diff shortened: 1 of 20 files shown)


cheeseng:
@exoego Sorry for attending late to this, I'll try to measure the gain and see how it is going.

or for Corporate:

Cheers.

exoego:
@cheeseng I sent a copy of signed CLA 📫

cheeseng:
@exoego Received with thanks!  I shall try to benchmark the difference in compile time ASAP.

cheeseng:
@exoego I just did a measurement with the following results in seconds: 

| Run | Without @noinline | With @noinline |
|-------|--------------------------|----------------------|
|     1 |                         416 |                   410 |
|     2 |                         415 |                   417 |
|     3 |                         415 |                   417 |
| Avg |                    415.33 |              414.67 |

The tests was done with scalatestTestJS/test:fastOptJS, and resulting .js files are both 338.1MB.

Unfortunately it looks not very encouraging, I suppose https://github.com/scala-js/scala-js/pull/2964 might have fixed the problem to some extend.

I wonder if the changes are only limit to those that takes => Any ?  Would it have any effect for other higher order function like => Unit (for scope level construct like 'describe' in FunSpec) or => Future[x] (for async suites)?  If it does we may consider adding @noinline for them and retry the measurement.

Fyi @sjrd @bvenners

plinlor:
@cla-bot[bot] check

artimasites:
@cla-bot[bot] check

exoego:
Closing since no clear benefits

PR #1758, opened 2020-01-14
minor source tweaks to allow building on JDK 11+

SethTisue:
use String#linesIterator, since Java stdlib now has String#lines


cheeseng:
@SethTisue @bvenners  I tested this on my machine with jdk 11.0.6, under jvm only one test failed in scalactic tests due to different error message from the jdk itself when array index out of bound happens, works well with js and native build, but dotty doesn't build, I guess dotty 0.21.0-RC1 probably does not support jdk 11 yet?

SethTisue:
> I guess dotty 0.21.0-RC1 probably does not support jdk 11 yet?

I don't know, but it wouldn't surprise me.

Issue #1767, opened 2020-02-04
invalid junit file format for property elements


LorenzMende:
Using the -u (Junit XML Reporter) with the scalatest Runner leads to invalid format error when integrating the result in Jenkins builds. (Windows, Jenkins 2.204.2, xUnit plugin 2.3.7).

> WARNING: The file 'xxx.xml' is an invalid file.
> WARNING: At line y of file xxx.xml:cvc-complex-type.2.1: Element 'property' darf kein Zeichen- oder Elementinformationselement [untergeordnete Elemente] haben, da der Contenttyp des Typs leer ist.

The warning/ error (result is skipped) can be translated to a xsd validation error of the property elements. Looking at the junit schema (either junit official or jenkins xunit plugin) the property elements do not have a type and therefor shall not contain further text content.

When looking at the scalatest/src/main/scala/org/scalatest/tools/JUnitXmlReporter.scala source, the function genPropertiesXml generates elements with empty text content.
A solution would be to correct the function.
I'll create a PR.

bvenners:
merged into 3.1.x ,will cherry pick to 3.2.x-new

PR #1768, opened 2020-02-04
[#1767] correcting text content of property elements of jUnit result files

LorenzMende:
* property elements are now non-text elements instead of empty-text
see `https://github.com/scalatest/scalatest/issues/1767` for more info.


cheeseng:
This looks good to me, and I think the fix should be cherry-picked into 3.2.x-new when it is merged into 3.1.x.

or for Corporate:

Cheers.

PR #1770, opened 2020-02-07
[3.1.x] Scala-js 1.0.0 Support

cheeseng:
Fixed failing tests for ObjectMeta when built with Scala-js 1.0.0, now all tests passed for Scala-js 1.0.0.

### DIFF
--- a/project/JsBuild.scala
+++ b/project/JsBuild.scala
@@ -15,7 +15,7 @@ import org.portablescala.sbtplatformdeps.PlatformDepsPlugin.autoImport._
 trait JsBuild { this: BuildCommons =>
 
   val scalaJSVersion = Option(System.getenv("SCALAJS_VERSION")).getOrElse("1.0.0")
-  val sjsPrefix = if (scalaJSVersion.startsWith("1.0.")) "_sjs1.0_" else "_sjs0.6_"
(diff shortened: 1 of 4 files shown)


project/JsBuild.scala
```diff
@@ -15,6 +15,7 @@ import org.portablescala.sbtplatformdeps.PlatformDepsPlugin.autoImport._
 trait JsBuild { this: BuildCommons =>
 
   val scalaJSVersion = Option(System.getenv("SCALAJS_VERSION")).getOrElse("1.0.0")
+  val sjsPrefix = if (scalaJSVersion.startsWith("1.0.")) "_sjs1.0_" else "_sjs0.6_"
```
sjrd: It's actually `"_sjs1_"` for 1.x.

sjrd: And it will stay like that for all 1.x version, so the test should be `startsWith("1.")`, not `startsWith("1.0.")`.

cheeseng: @sjrd Adjusted in the following commit: 

https://github.com/scalatest/scalatest/pull/1770/commits/4d6e31855a9c223e0534e33b948c0549e0f8a6fd

Thanks for catching it!

exoego:
Expecting this get released soon!
ScalaTest is widely adopted in Scala/Scala.js ecosystems.
It would allow the entire ecosystems moving forward to Scala.js 1.0.0.

PR #1774, opened 2020-02-11
Fix stray timer in SuiteSortingReporter & TestSortingReporter

cheeseng:
This is to address the same issue as this PR: 

https://github.com/scalatest/scalatest/pull/1591

In addition, we also added the missing shutdown calls to ExecutorService created in tests.

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/tools/SuiteSortingReporter.scala
+++ b/scalatest/src/main/scala/org/scalatest/tools/SuiteSortingReporter.scala
@@ -45,8 +45,7 @@ private[scalatest] class SuiteSortingReporter(dispatch: Reporter, val testSortin
     }
   }
   
-  private val timer = new Timer
-  private var timeoutTask: Option[TimeoutTask] = None
+  private var timeoutTask: Option[(TimeoutTask, Timer)] = None
 
   def doApply(event: Event): Unit = {
     synchronized {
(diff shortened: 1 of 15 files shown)


scalatest/src/main/scala/org/scalatest/tools/TestSortingReporter.scala
```diff
@@ -285,7 +288,8 @@ private[scalatest] class TestSortingReporter(suiteId: String, dispatch: Reporter
     synchronized {
       if (waitingBuffer.size > 0) {
         val head = waitingBuffer.head
-        if (timeoutTask.get.slot.uuid == head.uuid) { // Probably a double check, or just in case there's race condition
+        val (task, _) = timeoutTask.get
```
bvenners: This .get was there before, but I am not convinced the timeoutTask will always be defined. I'll keep looking at it.

scalatest/src/main/scala/org/scalatest/tools/SuiteSortingReporter.scala
```diff
@@ -283,23 +284,38 @@ private[scalatest] class SuiteSortingReporter(dispatch: Reporter, val testSortin
...
+      case None =>
+    }
+  }
   
   private def timeout(): Unit = {
     synchronized {
       if (slotListBuf.size > 0) {
         val head = slotListBuf.head
-        if (timeoutTask.get.slot.suiteId == head.suiteId) { // Probably a double check, or just in case there's race condition
+        val (task, _) = timeoutTask.get
```
bvenners: Here's the corresponding .get. Again not sure that it will always be defined.

PR #1794, opened 2020-03-16
Avoid polling Scala futures

raboof:
`Futures` contained a `futureValueImpl` which would poll the future, unnecessarily
increasing the running time of the test in the (common) case that the future
completes successfully.

`Futures` is implemented by `ScalaFutures` and `JavaFutures`, where `JavaFutures` had
an override for this method to avoid the polling.

This PR moves the `futureValueImpl` implementation to `ScalaFutures` (since it was
only used there anyway), and removes the need to poll.

This also removes (the need for) `FuturesSpec`, since that mainly tested the polling
mechanism that is now neither used in the Java implementation nor in the Scala one.
The actual logic is tested in `JavaFuturesSpec` and `ScalaFuturesSpec` respectively.


cheeseng:
@cla-bot[bot] check

cheeseng:
@raboof Looks good to me!

Do you mind to sign our CLA?

https://www.artima.com/cla/choose-type

Thanks!

artimasites:
@cla-bot[bot] check

raboof:
> @raboof Looks good to me!

Thanks!

> Do you mind to sign our CLA?

Done!

@cla-bot[bot] check

PR #1795, opened 2020-03-18
Alternative way to enable fill file path names in scalactic

saksmt:
Enabling this setting through environment variable isn't always available, so there is alternative way: property.

Example where it is not possible: bazel (and probably other monorepository build tools too)

P.S. Is there a way to backport this to 3.0.x?


cheeseng:
@cla-bot[bot] check

plinlor:
@cla-bot[bot] check

PR #1808, opened 2020-05-05
be Scala 2.13.3 friendly (Symbol#toString changed)

SethTisue:
the old code relied on this assumption:
`mySymbol.name == mySymbol.toString.drop(1)`
but Scala 2.13.3 is changing `Symbol#toString` to be
e.g. `Symbol(foo)` instead of `'foo`

the new code works fine on any Scala version.  I cannot see
any reason for the old code to be the way it was

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/PrivateMethodTester.scala
+++ b/scalatest/src/main/scala/org/scalatest/PrivateMethodTester.scala
@@ -160,7 +160,7 @@ trait PrivateMethodTester {
       import invocation._
 
       // If 'getMessage passed as methodName, methodNameToInvoke would be "getMessage"
-      val methodNameToInvoke = methodName.toString.substring(1)
+      val methodNameToInvoke = methodName.name
 
       def isMethodToInvoke(m: Method) = {
 


SethTisue:
This issue came up in the Scala community build. Not only did ScalaTest's own tests fail on 2.13.3-bin-43b70d4, but downstream projects were also affected.

I suggest that you roll a 3.1.x release with this fix sooner rather than later, in the hopes that by the time 2.13.3 is released and people upgrade to it, many people will already be on the new ScalaTest version and thus will never notice any problem.

SethTisue:
Not sure whether a 3.0.x backport-and-release is worth the trouble, or if you just want to tell 3.0.x users who hit this that they just ought to move to 3.1.x.

cheeseng:
@SethTisue I am looking into this now, thanks for the fix!

cheeseng:
I tested this fix with scala version 2.10-2.13, scala-js, and native build, seems to work correctly.

@bvenners Do you mind if we release a 3.1.2 for this?

bvenners:
@cheeseng Yes, let's do a 3.1.2 release with this fix.

SethTisue:
I think a number of your tests are also going to fail on 2.13.3; I see a number of places like:

```
    describe("a(Symbol) method returns Matcher") {
      val mt = be a ('file)
      
      it("should have pretty toString") {
        mt.toString should be ("be a 'file")
      }
```

where it now comes out as `be a Symbol(file)`

that's a lesser concern than the one I addressed in this PR, of course.

PR #1824, opened 2020-06-10
Use overload instead of override to implement Assertions.{assert, assume}

ohze:
+ Background
Because dotty don't permit `inline def` override `inline def`,  so in commit 3cfd618, in `Assertions.{assert, assume}`, we [pass](https://github.com/scalatest/scalatest/commit/3cfd6181bf1036ae046d269841b7518e0c35141a#diff-bf48f34ea897c0a7226a9493c8fb9789R471) `'this` to `AssertionsMacro`. And in `AssertionsMacro`, we [switch the implementation logic](https://github.com/scalatest/scalatest/commit/3cfd6181bf1036ae046d269841b7518e0c35141a#diff-547915506e8d905952154b3226aac560R33) of assert/ assume based on type of the passed param.

+ Problem
The solution above is not flexible enough if we want to [extends Assertions](https://github.com/scalatest/scalatestplus-junit/pull/12/files#diff-27382c1a71508ac8d6b28dc59afeb5a8R7) as in scalatest/scalatestplus-junit project.

+ Solution in this PR
As suggested in https://github.com/lampepfl/dotty/pull/8601#pullrequestreview-380646858, I use overload instead of override to implement `Assertions.{assert, assume}`

### DIFF
--- a/scalatest.dotty/src/main/scala/org/scalatest/Assertions.scala
+++ b/scalatest.dotty/src/main/scala/org/scalatest/Assertions.scala
@@ -468,7 +468,7 @@ trait Assertions extends TripleEquals  {
    * @throws TestFailedException if the condition is <code>false</code>.
    */
   inline def assert(inline condition: Boolean)(implicit prettifier: Prettifier, pos: source.Position): Assertion =
-    ${ AssertionsMacro.assert('this, 'condition, 'prettifier, 'pos, '{""}) }
+    ${ AssertionsMacro.assert('{condition}, '{prettifier}, '{pos}, '{""}) }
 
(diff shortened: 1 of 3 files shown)


cheeseng:
@ohze @bvenners Since 3.2.0 is around the corner, may I suggest we bring this dotty support to 3.2.0 instead of 3.1.x?

cheeseng:
@ohze Fyi I have cherry-pick your commit(s) into this branch here for the upcoming 3.2.0: 

https://github.com/scalatest/scalatest/tree/feature-3.2.0-with-dotty-improvements

cheeseng:
@ohze Fyi changes included in this new PR for 3.2.0 https://github.com/scalatest/scalatest/pull/1827

cheeseng:
@ohze Closing as this is now merged into 3.2.x-new and included in 3.2.0 release.

Issue #1788, opened 2020-03-03
private val in case class


charpov:
The code below fails with an exception `class org.scalactic.source.ObjectMeta$$anon$1 cannot access a member with modifiers "private"`.  I know that case classes with private fields are frown upon, but are they illegal?

```scala
case class S(private val v: Vector[Int]) {
  override def toString: String = v.mkString
}

object S {
  def apply(x: Int*): S = S(x.toVector)
}

class SampleTests extends AnyFunSuite {

  test("test") {
    val p1 = S(1,2,3)
    val p2 = S(2,3,4)
    assert(p2 == p1)
  }
}
```

bvenners:
That looks more like a bug, because you are not accessing the private member. Can you paste in the whole stack trace?

charpov:
Here's the stack trace:
```
class org.scalactic.source.ObjectMeta$$anon$1 cannot access a member of class S with modifiers "private"
java.lang.IllegalAccessException: class org.scalactic.source.ObjectMeta$$anon$1 cannot access a member of class S with modifiers "private"
	at java.base/jdk.internal.reflect.Reflection.newIllegalAccessException(Reflection.java:376)
	at java.base/java.lang.reflect.AccessibleObject.checkAccess(AccessibleObject.java:642)
	at java.base/java.lang.reflect.Method.invoke(Method.java:559)
	at org.scalactic.source.ObjectMeta$$anon$1.value(ObjectMeta.scala:54)
	at org.scalactic.ObjectDiffer.$anonfun$difference$7(Differ.scala:233)
	at scala.collection.StrictOptimizedIterableOps.flatMap(StrictOptimizedIterableOps.scala:118)
	at scala.collection.StrictOptimizedIterableOps.flatMap$(StrictOptimizedIterableOps.scala:105)
	at scala.collection.immutable.Vector.flatMap(Vector.scala:119)
	at org.scalactic.ObjectDiffer.difference(Differ.scala:232)
	at org.scalactic.ObjectDiffer.difference$(Differ.scala:225)
	at org.scalactic.ObjectDiffer$.difference(Differ.scala:274)
	at org.scalactic.AnyDiffer.difference(Differ.scala:287)
	at org.scalactic.Prettifier.apply(Prettifier.scala:130)
	at org.scalactic.Prettifier.apply$(Prettifier.scala:129)
	at org.scalactic.Prettifier$$anon$2.apply(Prettifier.scala:185)
[...]

PR #1834, opened 2020-06-25
Fix IllegalAccessException when comparing 2 instances of class has private fields

phucnh:
Fix an issue https://github.com/scalatest/scalatest/issues/1788

When comparing two instances of class which contains private fields, `IllegalAccessException: cannot access a member with modifiers "private"` occurs at ObjectMeta.
It caused because `Differ` try to access to the instance's value for generating differences analysis message.

I fixed this problem by setting the field accessible to true temporarily.

### DIFF
--- a/jvm/scalactic-test/src/test/scala/org/scalactic/source/ObjectMetaSpec.scala
+++ b/jvm/scalactic-test/src/test/scala/org/scalactic/source/ObjectMetaSpec.scala
@@ -21,7 +21,7 @@ class ObjectMetaSpec extends funspec.AnyFunSpec with matchers.should.Matchers {
 
   case class Person(name: String, private val age: Int) {
     val otherField = "test other field"
-    val privateField = "test private field"
+    private val privateField = "test private field"
   }
 
   describe("ObjectMeta") {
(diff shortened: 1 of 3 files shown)


artimasites:
@cla-bot[bot] check

jvm/scalactic-test/src/test/scala/org/scalactic/source/ObjectMetaSpec.scala
```diff
@@ -19,21 +19,23 @@ import org.scalatest._
 
 class ObjectMetaSpec extends funspec.AnyFunSpec with matchers.should.Matchers {
 
-  case class Person(name: String, age: Int) {
+  case class Person(name: String, private val age: Int) {
     val otherField = "test other field"
+    val privateField = "test private field"
```
cheeseng, line 24: @phucnh  Er, do you actually intended to use a private val here?

phucnh: @cheeseng Thank you for reviewing.
I have fixed https://github.com/scalatest/scalatest/pull/1834/commits/bb6a1e0bdd3a8b18e8cbb3efbc263d597ad785fe

cheeseng:
@phucnh There's a test on ObjectMetaSpec that is failing when built under Dotty, I have submitted the following PR to your branch to fix it: 

https://github.com/phucnh/scalatest/pull/1

It seems like the behavior in Dotty is preferred one as compared to Scala 2.

phucnh:
@cheeseng Thank you so much for fixing the failed test :bow:
I have merged your pull request

PR #1840, opened 2020-07-01
Fix composite status bundle succeeded

tim-zh:


cheeseng:
@tim-zh I submitted the following PR for your branch: 

https://github.com/tim-zh/scalatest/pull/1

If you undo your change the following test should actually fails: 

https://github.com/scalatest/scalatest/blob/3.1.x/scalatest-test/src/test/scala/org/scalatest/StatusSpec.scala#L343

It passed before because of another bug that I just fixed in https://github.com/tim-zh/scalatest/pull/1 , so let's fix both of them together.

Thanks.

cheeseng:
@bvenners I think this fix makes sense because you're using 'and' for processing the succeeded value in the bundle of the composite status: 

https://github.com/scalatest/scalatest/blob/3.1.x/scalatest/src/main/scala/org/scalatest/Status.scala#L1021

The original default value of 'false' will make the value always false, we actually have a test for it here: 

https://github.com/scalatest/scalatest/blob/3.1.x/scalatest-test/src/test/scala/org/scalatest/StatusSpec.scala#L343

but it seems to be negated by another bug we have which I fixed here: 

https://github.com/tim-zh/scalatest/pull/1

Do you mind to verify if both fix is correct?

Thanks!

artimasites:
@cla-bot[bot] check

tim-zh:
@cheeseng thanks!
sorry for late response

PR #1843, opened 2020-07-01
3.0: Keep suite reporter in map until the SuiteCompleted message is dealt with

johnduffell:
When running Async tests, scalatest buffers the output for all but the earliest test that's still running, outputting the remaining tests when the first test completes.
However there is a bug in that when all the tests in the suite are completed, and we fire the completed event, the reporter removes the  underlying reporter for the test.
This means when the SuiteCompleted event comes in later, there is no reporter to receive it and the test goes unreported (although the pass/fail status is still retained)  We found this was confusing when looking at our test results.
It looks like the removal from the map was a measure to save memory.

This PR moves the call to when the suite has really finished rather than just the tests inside the suite.  This seems to have fixed the bug in local testing.

### DIFF
--- a/scalatest/src/main/scala/org/scalatest/tools/SuiteSortingReporter.scala
+++ b/scalatest/src/main/scala/org/scalatest/tools/SuiteSortingReporter.scala
@@ -174,6 +174,7 @@ private[scalatest] class SuiteSortingReporter(dispatch: Reporter, val testSortin
           dispatch(doneEvent)
           dispatchToRegisteredSuiteReporter(head.suiteId, doneEvent)
         }
+        suiteReporterMap -= (head.suiteId)
         slotListBuf = fireReadySuiteEvents(slotListBuf.tail)
         if (slotListBuf.size > 0) 
           scheduleTimeoutTask()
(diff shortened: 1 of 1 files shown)


artimasites:
@cla-bot[bot] check

scalatest/src/main/scala/org/scalatest/tools/SuiteSortingReporter.scala
```diff
@@ -174,6 +174,7 @@ private[scalatest] class SuiteSortingReporter(dispatch: Reporter, val testSortin
           dispatch(doneEvent)
           dispatchToRegisteredSuiteReporter(head.suiteId, doneEvent)
         }
+        suiteReporterMap -= (head.suiteId)
```
cheeseng, line 177: @johnduffell I am not sure if this is needed, it seems like you have also added the same code in the fireReadySuiteEvents which is called right after this line, perhaps I am missing something here?

johnduffell: `fireReadySuiteEvents` is only called with the tail of the `slotListBuf`, we are only dealing with the head by the looks of it in this function.
I may well have not fully understood the code, but it looked like I needed to cover both.

cheeseng: @johnduffell I see, I think you're right, I was missing the .tail part.  I'll cherry-pick this commit into 3.1.x as well.

Thanks!

PR #1844, opened 2020-07-01
3.2: Keep suite reporter in map until the SuiteCompleted message is dealt with

johnduffell:
This is a forward port of https://github.com/scalatest/scalatest/pull/1843
The same issue is occurring in 3.2, when running two trivial ASync tests from SBT only one of the tests is reported, the other is not because the underlying Reporter is being removed when all of the child tests are complete, and not waiting for the SuiteCompleted event.  This means that when the second test is flushed to the reporter, there is no reporter, so the output is lost (although the failing test is still failing)

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/tools/SuiteSortingReporter.scala
+++ b/jvm/core/src/main/scala/org/scalatest/tools/SuiteSortingReporter.scala
@@ -173,6 +173,7 @@ private[scalatest] class SuiteSortingReporter(dispatch: Reporter, val testSortin
           dispatch(doneEvent)
           dispatchToRegisteredSuiteReporter(head.suiteId, doneEvent)
         }
+        suiteReporterMap -= (head.suiteId)
         slotListBuf = fireReadySuiteEvents(slotListBuf.tail)
         if (slotListBuf.size > 0) 
           scheduleTimeoutTask()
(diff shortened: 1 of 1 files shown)


artimasites:
@cla-bot[bot] check

jvm/core/src/main/scala/org/scalatest/tools/SuiteSortingReporter.scala
```diff
@@ -173,6 +173,7 @@ private[scalatest] class SuiteSortingReporter(dispatch: Reporter, val testSortin
           dispatch(doneEvent)
           dispatchToRegisteredSuiteReporter(head.suiteId, doneEvent)
         }
+        suiteReporterMap -= (head.suiteId)
```
cheeseng, line 176: @johnduffell I am not sure if this is needed, it seems like you have also added the same code in the fireReadySuiteEvents which is called right after this line, perhaps I am missing something here?

johnduffell: same comment as https://github.com/scalatest/scalatest/pull/1843#pullrequestreview-451555737
> `fireReadySuiteEvents` is only called with the tail of the `slotListBuf`, we are only dealing with the head by the looks of it in this function.
> I may well have not fully understood the code, but it looked like I needed to cover both.

PR #1861, opened 2020-07-29
[3.2] Avoid polling Scala futures

cheeseng:
This is ported from 3.1.x which includes changes from the following PR: 

https://github.com/scalatest/scalatest/pull/1794
https://github.com/scalatest/scalatest/pull/1860


raboof:
Thanks!

PR #1869, opened 2020-08-04
Allow pendingUntilFixed to catch Throwables

disordered:
Some frameworks (ZIO, for example) may result in an error that extends from `Throwable` directly instead of an `Exception` when effect is run. Test code then has to be wrapped into an extra try/catch to mark test pending.

If you think catching `Throwable` is too drastic due to catching fatal errors too, we could add extra patterns for fatal errors, or allow all subtypes of `Error` to propagate.

### DIFF
--- a/dotty/core/src/main/scala/org/scalatest/Assertions.scala
+++ b/dotty/core/src/main/scala/org/scalatest/Assertions.scala
@@ -1263,7 +1263,7 @@ trait Assertions extends TripleEquals  {
    * </p>
    *
    * @param f a block of code, which if it completes abruptly, should trigger a <code>TestPendingException</code>
-   * @throws TestPendingException if the passed block of code completes abruptly with an <code>Exception</code> or <code>AssertionError</code>
+   * @throws TestPendingException if the passed block of code completes abruptly with a <code>Throwable</code>
    */
   def pendingUntilFixed(f: => Unit)(implicit pos: source.Position): Assertion with PendingStatement = {
     val isPending =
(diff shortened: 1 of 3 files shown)


disordered:
> Hi @disordered, we require contributors to sign our Contributor License Agreement, and we don't have yours on file. In order for us to review and merge your code, please access https://www.artima.com/cla/choose-type to sign our Contributor License Agreement. Your effort is highly appreciated. Thank you.

I reviewed the agreement, and I don't feel like sharing my contact information and address with you guys. You can still use the contribution as you wish, though.

bvenners:
@disordered Hmm, curious why does ZIO throw exceptions that Throwable? That seems unidiomatic on the JVM.

disordered:
> @disordered Hmm, curious why does ZIO throw exceptions that Throwable? That seems unidiomatic on the JVM.

I don't really know. I would have guessed that it is their way to suppress stacktraces (ZIO tracks fiber failures accross threads), but turns out you can suppress stacktrace on exceptions too.

Maybe scaladoc on `FiberFailure` sheds some light:

> Represents a failure in a fiber. This could be caused by some non- recoverable error, such as a defect or system error, by some typed error, or by interruption (or combinations of all of the above).
>This class is used to wrap ZIO failures into something that can be thrown, to better integrate with Scala exception handling.

disordered:
@bvenners I switched the catch to [NonFatal](https://www.scala-lang.org/api/2.13.3/scala/util/control/NonFatal$.html ) instead, which will not catch what scala considers fatal throwables.

PR #1874, opened 2020-08-08
Quote path for regex

martijnhoekstra:
Windows paths contain backslash which will be interpreted by the regex engine if left unquoted.

Targetting 3.1.x. Should it? Or should this be retargeted?


artimasites:
@cla-bot[bot] check

PR #1878, opened 2020-08-13
3.2.1 Fix

cheeseng:
- Remove suite reporter only when the suite's tests are completed.
- Minor tweak to futureValueImpl to skip Await.ready when future is already available at the call site.

@johnduffell Do you mind to sanity check if this still fix your issue intended to be fixed in https://github.com/scalatest/scalatest/pull/1844 ?

Cheers.


johnduffell:
Sorry to hear about the regression.  I have a task to upgrade our project soon so I will check both  3.2.1 and 3.2.2-M1 and let you know.
I didn't add any tests for these scenarios, perhaps the existing tests do not cover this area.

cheeseng:
@johnduffell Sorry I have not able to deployed 3.2.2-M1 today as try to look into another issue of futureValueImpl, I shall attempt again tomorrow.

Issue #1883, opened 2020-08-28
Reporter completed abruptly with TryValues


seblm:
Since #1738 there is a regression:
given `SocketReporter` is registered (`-K host port`)
when any assertion failure is raised by `TryValues`
then `SocketReporter` completed abruptly

```
Exception in thread "Thread-37" java.io.WriteAbortedException: writing aborted; java.io.NotSerializableException: org.scalatest.TryValues$SuccessOrFailure
	at java.io.ObjectInputStream.readObject0(ObjectInputStream.java:1654)
...
Caused by: java.io.NotSerializableException: org.scalatest.TryValues$SuccessOrFailure
	at java.io.ObjectOutputStream.writeObject0(ObjectOutputStream.java:1184)
...
Reporter completed abruptly with an exception after receiving event: TestFailed(Ordinal(0, 4),The Try on which success was invoked was not a Success; it was Failure(java.lang.Exception: failure).,MySpec,package.MySpec,Some(package.MySpec),My should works,should works,Vector(),Vector(),Some(org.scalatest.exceptions.TestFailedException: The Try on which success was invoked was not a Success; it was Failure(java.lang.Exception: failure).),Some(519),Some(IndentedText(- should works,should works,1)),Some(SeeStackDepthException),Some(package.MySpec),None,pool-1-thread-1-ScalaTest-running-MySpec,1598614861046).
java.io.NotSerializableException: org.scalatest.TryValues$SuccessOrFailure
```

seblm:
Here is [a minimal project](https://github.com/seblm/try-values-not-serializable) that reproduce the defect.

seblm:
For the record, #1738 introduces the regression onto [3.1.x branch](../tree/3.1.x) but it is actually #1878 that cherry-pick this change to [3.2.x branch](..).

kpbochenek:
I am also getting something similar:
```
[...]

PR #1884, opened 2020-08-30
Avoid NotSerializableException when TryValues or EitherValues raise assertion failure

seblm:
Fixes scalatest/scalatest#1883

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/concurrent/AbstractPatienceConfiguration.scala
+++ b/jvm/core/src/main/scala/org/scalatest/concurrent/AbstractPatienceConfiguration.scala
@@ -88,10 +88,31 @@ trait AbstractPatienceConfiguration extends ScaledTimeSpans {
    * @author Bill Venners
    * @author Chua Chee Seng
    */
-  final case class PatienceConfig(timeout: Span = scaled(Span(150, Millis)), interval: Span = scaled(Span(15, Millis)))
+  final case class PatienceConfig(timeout: Span = scaled(Span(150, Millis)), interval: Span = scaled(Span(15, Millis))) {
+    /**
+     * <code>PatienceConfig</code> is an inner class. For some reason (like using <code>SocketReporter</code> and having
+     * a <code>ScalaFutures</code> test that fails), instance of this class will be serialized. But as an inner class,
+     * scala will add an extra <code>$outer</code> field that references actual outer class: an instance that you don't
+     * want make serializable.
+     * To avoid errors like <code>java.io.NotSerializableException: org.scalatest.verbs.BehaveWord</code>, this function
+     * delegates serialization to a non inner class.
+     */
+    private def writeReplace(): Any = new AbstractPatienceConfiguration.PatienceConfigProxy(timeout, interval)
+  }
 
   /**
    * Returns a <code>PatienceConfig</code> value providing default configuration values if implemented and made implicit in subtraits.
    */
   def patienceConfig: PatienceConfig
+}
+
+private object AbstractPatienceConfiguration extends AbstractPatienceConfiguration {
+
+  override def patienceConfig: PatienceConfig = PatienceConfig()
+
+  @SerialVersionUID(3466997156436169627L)
+  private final class PatienceConfigProxy(timeout: Span, interval: Span) extends Serializable {
+    private def readResolve(): Any = PatienceConfig(timeout, interval)
+  }
+
 }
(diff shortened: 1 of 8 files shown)


artimasites:
@cla-bot[bot] check

kpbochenek:
@seblm if you add custom serialization shouldn't you also add custom deserializer? Won't there be problems on side receiving this event? 

Moving `PatienceConfiguration` to not be inner class would be too big change right?

seblm:
> @seblm if you add custom serialization shouldn't you also add custom deserializer? Won't there be problems on side receiving this event?
You've right: serializing then deserializing produce an error. I'll fix it.

> Moving `PatienceConfiguration` to not be inner class would be too big change right?
Unfortunately, this change will not be binary compatible. I think it should be introduced in a major or a minor version instead of patch version.

seblm:
> > @seblm if you add custom serialization shouldn't you also add custom deserializer? Won't there be problems on side receiving this event?
> 
> You've right: serializing then deserializing produce an error. I'll fix it.

With a proxy which is not an inner class, serialization of `PatienceConfig` is ok. @kpbochenek you can have a look.

kpbochenek:
> > > @seblm if you add custom serialization shouldn't you also add custom deserializer? Won't there be problems on side receiving this event?
> > 
> > 
> > You've right: serializing then deserializing produce an error. I'll fix it.
> 
> With a proxy which is not an inner class, serialization of `PatienceConfig` is ok. @kpbochenek you can have a look.

I don't know much about java serialization, as long as it works :+1:

Just hope it can be merged and released quickly :pray:

cheeseng:
@seblm @kpbochenek Sorry for the late response, has been away for other works.  I am looking into this  now.

cheeseng:
@seblm @kpbochenek This PR unfortunately is having problem under scala-js and dotty build, I have submitted the following PR to your clone: 

https://github.com/seblm/scalatest/pull/1

Can you sanity check if the fix works for you?

Note that I have run MIMA test for this PR + New PR, seems to pass.  Hopefully we can release a new release containing this fix soon.

kpbochenek:
Thanks for looking into it!
I will try to find a time and try it but it might not be that easy, tests show problem very well so if they pass I think that should be enough :+1:

seblm:
Thanks a lot @cheeseng for your pull request. It is now rebased onto my fork.
I was not aware of scala-js and dotty exceptions.

cheeseng:
@bvenners This PR looks good to me now, and it passed MIMA test, do you mind to review for merging when it is convenient for you?  Do you mind if we release 3.2.3 and 3.1.5 for this?

Cheers.

jvm/core/src/main/scala/org/scalatest/concurrent/AbstractPatienceConfiguration.scala
```diff
@@ -88,10 +89,31 @@ trait AbstractPatienceConfiguration extends ScaledTimeSpans {
...
+  }
 
   /**
    * Returns a <code>PatienceConfig</code> value providing default configuration values if implemented and made implicit in subtraits.
    */
   def patienceConfig: PatienceConfig
 }
+
+protected[concurrent] object AbstractPatienceConfiguration extends AbstractPatienceConfiguration {
+  override def patienceConfig: PatienceConfig = PatienceConfig()
```
bvenners, line 111: Why is this protected[concurrent] instead of just private[concurrent]?

seblm: You've right: this visibility can be reduced to `private`. The change is applied.

bvenners: Tnx. Testing this now.

bvenners:
Merged. Thanks very much for this PR!

seblm:
Thank you to have merge it so quickly.

Do you think that it could be possible to drop jdk object serialization usage from scalatest?

bvenners:
@seblm Say you added SerialVersionUID annotations to two classes in this PR. Can you elaborate on why you thought that was important? I would think we may not want to give those, so they'll be generated from the class interface. Otherwise I think we have to pay attention to when we change it incompatibly, which we would likely miss. Tnx.

seblm:
@bvenners you have right on this: I added `SerialVersionUID` more by habit than by real conviction. It is ok to remove these annotations for the sake of maintainability.

seblm:
I am glad to see that #2086 will handle this problem more deeply 👍

PR #1902, opened 2020-10-08
Introduced support of specified version of Scala-Native

catap:
Scala-Native team made a huge job and right now it is possibly to build scala-2.12 application on scala-native.

This isn't released feature, but `-SNAPSHOT` version that's supported it.

Anyway, in near future (I hope) scalatest needs to add a new version of SN because current master isn't binary compatibly with `0.4.0-M2`.

This PR:
 - introduced support of specified Scala-Native version via environment variables;
 - cleaned up `NativeBuild.scala` to support both versions

Scala-native migrated to annotation based way to load unit tests, like Scala-js does: https://github.com/scala-native/scala-native/pull/1802.

I've added a commit to enabled it but with name `scala.scalajs.reflect.annotation.EnableReflectiveInstantiation` that is fine for both version `-M2` and `-SNAPSHOT`, and when the next version is released it should be renamed to `scala.scalanative.reflect.annotation.EnableReflectiveInstantiation`.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/Suites.scala
+++ b/jvm/core/src/main/scala/org/scalatest/Suites.scala
@@ -53,6 +53,7 @@ import org.scalactic.exceptions.NullArgumentException
  * @author Bill Venners
  */
 //SCALATESTJS-ONLY @scala.scalajs.reflect.annotation.EnableReflectiveInstantiation
+//SCALATESTNATIVE-ONLY @scala.scalajs.reflect.annotation.EnableReflectiveInstantiation
 class Suites(suitesToNest: Suite*) extends Suite { thisSuite =>
 
   requireNonNull(suitesToNest)
--- a/jvm/featurespec/src/main/scala/org/scalatest/featurespec/AnyFeatureSpecLike.scala
+++ b/jvm/featurespec/src/main/scala/org/scalatest/featurespec/AnyFeatureSpecLike.scala
@@ -43,6 +43,7 @@ import org.scalatest.exceptions._
  */
 @Finders(Array("org.scalatest.finders.FeatureSpecFinder"))
 //SCALATESTJS-ONLY @scala.scalajs.reflect.annotation.EnableReflectiveInstantiation
+//SCALATESTNATIVE-ONLY @scala.scalajs.reflect.annotation.EnableReflectiveInstantiation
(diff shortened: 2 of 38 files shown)


catap:
Ok, it works.

cheeseng:
@catap Changes looks good!  I'll try to build them locally and see how it goes.

catap:
@cheeseng yes, you need it.

As far as I know `-SNAPSHOT` version of scala-native doesn't publish anywhere.

So, you need to use the last master and run `sbt +publishLocal` that builds and publish locally version for 2.11 and 2.12.

cheeseng:
@catap Yes it builds after I built SN master locally with sbt +publishLocal, however, many tests now failing with error like this: 

```
Cause: java.lang.ClassCastException: org.scalactic.anyvals.NonZeroLong cannot be cast to java.lang.Long
[info]     at java.lang.Throwable.fillInStackTrace(Unknown Source)
[info]     at scala.scalanative.runtime.package$.throwClassCast(Unknown Source)
[info]     at <none>.(Unknown Source)
[info]     at org.scalactic.anyvals.NonZeroLongSpec$$Lambda$69.apply(Unknown Source)
[info]     at org.scalatest.enablers.UnitPropCheckerAsserting$PropCheckerAssertingImpl.$anonfun$checkForAll$1(Unknown Source)
[info]     at org.scalatest.enablers.UnitPropCheckerAsserting$PropCheckerAssertingImpl$$Lambda$7.apply(Unknown Source)
```

They used to pass in 0.4.0-M2, not sure what's failing them now.

Anyway, after changes in PR it still by default is building with 0.4.0-M2 (for now), so it looks good to me to merge this into 3.2.x-new.

catap:
@cheeseng I guess I had used `-M2` as default value. Haven't I?

catap:
@cheeseng interesting. I can reproduce it only via 2.12 version.

Looks like a bug inside SN. I open an issue: https://github.com/scala-native/scala-native/issues/1950

cheeseng:
@catap The last round I tested it, it actually happens on both 2.11.12 and 2.12.12, when using SN 0.4.0-SNAPSHOT that I built locally.

cheeseng:
@catap I'll continue the conversion on your report in SN https://github.com/scala-native/scala-native/issues/1950

PR #1917, opened 2020-11-03
Shrink With Rose Tree

cheeseng:
First step to try shrink with RoseTree.


bvenners:
I made a similar enhancement to feature-rose-tree-after-net directly. Going to close this one.

PR #1922, opened 2020-11-16
Fixed building scalatests on scala-native-2.13

catap:
This commit fixed a future issue with scala-native that can be discovered inside current scala-native master because it supports scala-2.13 :)

### DIFF
--- a/project/GenScalaTestNative.scala
+++ b/project/GenScalaTestNative.scala
@@ -116,10 +116,16 @@ object GenScalaTestNative {
   val genScalaPackages: Map[String, List[String]] =
     Map(
       "org/scalatest" -> (List(
+        "Doc.scala",
+        "DocSpec.scala",
+        "DocSpecLike.scala",
         "DispatchReporter.scala",
         "ConfigMapWrapperSuite.scala",    // skipped because depends on java reflection.
         "JavaClassesWrappers.scala",
         "Shell.scala",
+        "StreamlinedXml.scala",
+        "StreamlinedXmlEquality.scala",
+        "StreamlinedXmlNormMethods.scala",
         "SuiteRerunner.scala",
         "SuiteRerunner.scala",
         "run.scala"
--- a/project/NativeBuild.scala
+++ b/project/NativeBuild.scala
@@ -16,18 +16,6 @@ trait NativeBuild { this: BuildCommons =>
 
   val scalaNativeVersion = Option(System.getenv("SCALANATIVE_VERSION")).getOrElse("0.4.0-M2")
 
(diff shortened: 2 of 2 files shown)


cheeseng:
@catap Er, I wonder is there no plan to support scala-xml in scala-native?  scala-js does seem supports it in release 1.3.0 and 2.0.0-M2 for scala-js 0.6 and 1.

catap:
@cheeseng at least it hasn't got scala-xml for 1.0.6 :)

catap:
@cheeseng BTW I haven't see `scala-xml` for Scala-Native here: https://repo1.maven.org/maven2/org/scala-lang/modules/

:(

cheeseng:
@catap Ha, actually for 2.13 I can't find scala-native on maven repo, too.  :)

catap:
@cheeseng scala native hasn't yet released and 2.12 and 2.13 are available only from master as snapshot ;)

cheeseng:
@catap Not seeing it being included in scala-xml build: 

https://github.com/scala/scala-xml/blob/master/build.sbt#L164

May be first step will be to try add native build to scala-xml, not sure how hard it will be though.

catap:
@cheeseng so, if it hasn't included how does it works? :)

catap:
@cheeseng Looks like it never was includes: https://github.com/scala/scala-xml/issues/282

So, I don't understand how it works right now.

cheeseng:
@catap It seems like the reason they dropped SN support is due to it requires scala 2.12, which SN did not support, but now 2.12 support is coming in master, right?

catap:
@cheeseng and 2.13 also. But I have no idea when the master will be released.

Here is the loop: I'm using quite complex project (its unit test) to find a bug inside scala-native master to help the team to release 2.12/2.13 without bugs [you can check the issue, I did find a few deep things, and a few inisde SJS by this code base].

All this unit tests are implemented on scalatests and right now I'm in the loop.

But I feel that start to broke it from adding support of SN to scala-xml is good idea.

catap:
@cheeseng I've touched scala-xml a bit and it seems isn't easy to make optional SN support.

Special that a new SN is planned to release on this year I'd like to wait until it is released :)

I will back to this PR ASAP

SethTisue:
> is there no plan to support scala-xml in scala-native? 

https://github.com/scala/scala-xml/issues/282

catap:
@SethTisue until SN is released with support 2.12 and 2.13 it is impossibly.

I've tried make cross compilation with optionally enabled SN, but it is quite a challenge that doesn't seems worth to be solved because the release should be in months.

catap:
This also quite old and seem to not be needed.

@cheeseng shall I close it?

PR #1926, opened 2020-11-25
Upgrade to scala-3.0.0-M2

catap:


cheeseng:
@catap Thanks for the PR!  The change to project/plugins.sbt will complement the other similar PR here: 

https://github.com/scalatest/scalatest/pull/1927

PR #1928, opened 2020-11-30
.jvmopts: Increase max heap size to 11 GB

smarter:
The previous default of 4 GB lead to the JVM getting stuck in garbage
collection seemingly forever.


catap:
@smarter I feel déjà vu: https://github.com/scalatest/scalatest/pull/1904

:)

PR #1923, opened 2020-11-19
Upgrade to Scala 2.13.4

smarter:
This requires a single source code change to handle
https://github.com/scala/scala/pull/9292.

PR #1929, opened 2020-12-01
allow building on JDK 15

SethTisue:
this came up in the Scala community build, where we build on 11 and 15 as well as 8

### DIFF
--- a/.jvmopts
+++ b/.jvmopts
@@ -1,8 +1,5 @@
--server 
 -Xms512M 
 -Xmx4000M 
 -Xss10M
--XX:+CMSClassUnloadingEnabled 
--XX:+UseConcMarkSweepGC 
 -XX:NewRatio=8
--XX:MaxPermSize=512M
+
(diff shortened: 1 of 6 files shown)


.jvmopts
```diff
@@ -1,8 +1,5 @@
--server 
 -Xms512M 
 -Xmx4000M
```
catap, line 2: I suggest to increase heap because it isn't enough to build SN.

See example of such attempts at https://github.com/scalatest/scalatest/pull/1904 or https://github.com/scalatest/scalatest/pull/1929

SethTisue: this is 1929, you meant 1904 and...?

SethTisue: oh, I see it now, #1928

catap: @SethTisue and this is that only I know!

jvm/refspec/src/main/scala/org/scalatest/refspec/RefSpec.scala
```diff
@@ -1383,7 +1383,7 @@ private[scalatest] object RefSpec {
         else
           value
           
-      val bytes = Codec.toUTF8(toHash.toArray)
+      val bytes = Codec.toUTF8(toHash)
```
smarter, line 1386: This change is already part of https://github.com/scalatest/scalatest/pull/1923

PR #1936, opened 2020-12-11
Inside's Position for Scala 3

cheeseng:
In Scala 2, in the following code that uses ScalaTest's 'Inside' feature: 

```
inside(rec) {
  case Record(name, address, age) =>
     ...
}
```

When the rec passed in does not match any case in the partial function, an error is reported at line 1 where the inside(rec) is called, the line position is done through Position.here macro expansion like this: 

```
inside(rec) {
  case Record(name, address, age) =>
     ...
}(Position.here)
```

In Scala 2 macro, the Position.here returns line 1 where inside(rec) is called, which is our desired behavior.  However, in Scala 3's Position.here implementation, it actually reports the position of line 4 where the Position.here happens.

Technically, I think Scala 3 is the correct behavior as that's where the macro expansion happened, so in this PR I am trying to change the inside function (for Scala 3 version) itself a macro expansion so we can get the position at line 1.

CC @smarter @liufengyun , is there a better alternative for the problem?


liufengyun:
@cheeseng The approach you take makes sense: I don't see a solution without making `inside` a macro itself.

PR #1940, opened 2020-12-15
Generator Refactor Batch 1

cheeseng:
First Batch of Generator Refactor

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
+++ b/jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
@@ -1121,6 +1121,23 @@ object Generator {
     */
   implicit val posZLongGenerator: Generator[PosZLong] =
     new Generator[PosZLong] {
+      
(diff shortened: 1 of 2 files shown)


jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
```diff
@@ -850,77 +846,73 @@ object Generator {
...
+            }
+            else {
+              val sqrt: Float = math.sqrt(f.abs.toDouble).toFloat
+              if (sqrt < 1.0f) Rose(0.0f) :: acc
+              else {
+                val whole: Float = sqrt.floor
+                val negWhole: Float = math.rint((-whole).toDouble).toFloat
+                val (first, second) = if (f > 0.0f) (negWhole, whole) else (whole, negWhole)
+                shrinkLoop(first, Rose(first) :: Rose(second) :: acc)
+                }
```
bvenners, line 894: Oh I think these Roses needed to be RoseTrees also. I'll check it.

PR #1941, opened 2020-12-16
Generator Refactor Batch 2

cheeseng:
Second batch of generator refactoring.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
+++ b/jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
@@ -1249,6 +1249,34 @@ object Generator {
     */
   implicit val finiteFloatGenerator: Generator[FiniteFloat] =
     new Generator[FiniteFloat] {
+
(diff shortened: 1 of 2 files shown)


jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
```diff
@@ -634,41 +634,36 @@ object Generator {
...
 
+      case class NextRoseTree(value: Byte) extends RoseTree[Byte] {
+        def shrinks(rndPassedToShrinks: Randomizer): (List[RoseTree[Byte]], Randomizer) = {
+          def shrinkLoop(n: Byte, acc: List[RoseTree[Byte]]): List[RoseTree[Byte]] = {
+            if (n == 0) acc
+            else {
+              val half: Byte = (n / 2).toByte
+              if (half == 0) Rose(0.toByte) :: acc
+              else shrinkLoop(half, Rose((-half).toByte) :: Rose(half) :: acc)
+            }
```
bvenners, line 645: I think these two Roses should be RoseTrees. I'll fix that when I mere.

PR #1942, opened 2020-12-17
Generator Refactor Batch 3

cheeseng:
Third batch of generator refactoring.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
+++ b/jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
@@ -1954,6 +1954,33 @@ object Generator {
     */
   implicit val negDoubleGenerator: Generator[NegDouble] =
     new Generator[NegDouble] {
+
+      case class NextRoseTree(value: NegDouble) extends RoseTree[NegDouble] {
+        def shrinks(rndPassedToShrinks: Randomizer): (List[RoseTree[NegDouble]], Randomizer) = {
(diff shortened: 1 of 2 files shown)


jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
```diff
@@ -1357,56 +1353,52 @@ object Generator {
...
+                  Float.MaxValue
+                else fv
+              // Nearest whole numbers closer to zero
+              val nearest = PosZFloat.ensuringValid(n.floor)
+              shrinkLoop(nearest, NextRoseTree(nearest) :: acc)
+            }
+            else {
+              val sqrt: Float = math.sqrt(fv.toDouble).toFloat
+              if (sqrt < 1.0f) NextRoseTree(PosZFloat(0.0f)) :: acc
+              else {
```
bvenners, line 1376: @cheeseng I think if if we are not recursively calling shrinkLoop we should probably use Rose, because we are at the end of the line. I'm not 100% sure, but let's stick with that pattern unless we get evidence to the contrary. We need to make sure that shrinking terminates, so at some point there needs to be a node that has no subnodes. That's what the Rose is. I'll make such changes in this PR before merging.

cheeseng: @bvenners Noted, I'll make the related changes and include them in the next PR.

Thanks.

PR #1949, opened 2020-12-22
Enhancements for Scala 3

cheeseng:
- Replaced AnyShouldWrapper with extension methods in Scala 3.
- Got 'should compile' syntax to work.
- Made Eventually's eventually methods as macro in Scala 3, to fix the position problem, same approach as Inside.
- Made Futures's whenReady methods as macro in Scala 3, to fix the position problem, same approach as Inside.
- Enabled several test classes under org.scalatest.concurrent.


jvm/core/src/main/scala/org/scalatest/concurrent/Eventually.scala
```diff
@@ -438,4 +454,20 @@ trait Eventually extends PatienceConfiguration {
...
  * </pre>
  */
-object Eventually extends Eventually
+object Eventually extends Eventually {
+
+  //DOTTY-ONLY def callRetry[T](retrying: Retrying[T], timeout: Span, interval: Span, pos: source.Position, fun: => T): T = 
+  //DOTTY-ONLY   retrying.retry(timeout, interval, pos)(fun)
+
+  //DOTTY-ONLY import scala.quoted._
+  //DOTTY-ONLY private[scalatest] def eventuallyMacro[T](timeout: Expr[Span], interval: Expr[Span], fun: Expr[T], retrying: Expr[Retrying[T]])(using quotes: Quotes, typeT: Type[T]): Expr[T] = {
```
cheeseng, line 463: @liufengyun I am trying to use fun: Expr[=> T] here but unable to, do you mind to help us sanity check if using fun: Expr[T] is the right thing to do?

Thanks.

liufengyun: Yes, I think it's correct to use `Expr[T]` here. In macros, `Expr[=>T]` and `Expr[T]` don't have much difference.

cheeseng: @liufengyun Thanks for checking!

jvm/core/src/main/scala/org/scalatest/concurrent/Futures.scala
```diff
@@ -600,9 +605,37 @@ trait Futures extends PatienceConfiguration {
...
   }
-}
 
+  //DOTTY-ONLY import scala.quoted._
+
+  //DOTTY-ONLY final def workaroundWhenReadyImpl[T, U](futures: Futures)(future: Any, fun: Any, timeout: Span, interval: Span, pos: source.Position): U = 
+  //DOTTY-ONLY   whenReadyImpl(futures)(future.asInstanceOf[futures.FutureConcept[T]], fun.asInstanceOf[T => U], timeout, interval, pos)
+
+  //DOTTY-ONLY // Ideally, we can use future: Expr[futures.FutureConcept[T]] and fun Expr[T => U] here, can't get it to work so we have the above workaroundWhenReadyImpl that takes Any.
+  //DOTTY-ONLY private[concurrent] def whenReadyMacro[T, U](futures: Expr[Futures], future: Expr[Any], fun: Expr[Any], timeout: Expr[Span], interval: Expr[Span])(using quotes: Quotes): Expr[U] = {
```
cheeseng, line 631: @liufengyun I am trying to use a dependent type here for future argument, which I tried future: Expr[futures.FutureConcept] but it did't work.  Also for the 'fun' argument it is supposed to be Expr[T => U] but it didn't work also, is there a correct way to make dependent type to work?

Cheers.

liufengyun: @cheeseng You need `inline def whenReady[T: Type, U: Type]` in the macro definition for `Expr[T => U]` to work.

Path-dependent types like `Expr[futures.FutureConcept]` will not work, as the parameter `futures` has the type `Expr[Futures]` instead of `Futures`. However, you can use `Expr[Futures#FutureConcept]`, which is a little better than `Expr[Any]`. I don't see how to do better than this.

dotty/matchers-core/src/main/scala/org/scalatest/matchers/CompileMacro.scala
```diff
@@ -25,12 +25,12 @@ import scala.quoted._
...
 
   // check that a code snippet compiles
-  def assertCompileImpl[T](self: Expr[T], compileWord: Expr[CompileWord], pos: Expr[source.Position])(shouldOrMust: String)(using Quotes): Expr[Assertion] = {
+  def assertCompileImpl[T](self: Expr[T], typeChecked: Expr[Boolean], compileWord: Expr[CompileWord], pos: Expr[source.Position])(shouldOrMust: String)(using Quotes): Expr[Assertion] = {
     import quotes.reflect._
 
     // parse and type check a code snippet, generate code to throw TestFailedException if both parse and type check succeeded
     def checkCompile(code: String): Expr[Assertion] =
-      if (/*typeChecks(code)*/ true) '{ Succeeded } // FIXME
+      if (typeChecked.valueOrError) '{ Succeeded } // FIXME
```
bvenners, line 33: @cheeseng If this was fixed, then next round please delete the FIXME comment. Tnx.

cheeseng: @bvenners Sure, I'll do in the next-batch branch that I am working on currently.

Thanks.

PR #1962, opened 2021-01-28
Retrying's Scheduler Shutdown

cheeseng:
Rewritten Retrying.scala to do away with lazy val scheduler in object Retrying which is not shutdown, replaced with approach where new instance of scheduler is created every time and .shutdown() is called right after the intended task is scheduled with it, this way the scheduler will shutdown after the scheduled task completes its execution.


francisdb (approved):
Fixes the issue for me

PR #1966, opened 2021-02-18
3.2.4 Release

cheeseng:
- Further improved support for Scala 3.
- Bumped up to Scala 3.0.0-RC1.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/concurrent/Futures.scala
+++ b/jvm/core/src/main/scala/org/scalatest/concurrent/Futures.scala
@@ -472,7 +472,10 @@ trait Futures extends PatienceConfiguration {
      * @throws TestFailedException if the future is cancelled, expires, or is still not ready after
      *     the specified timeout has been exceeded
      */
-    def futureValue(implicit config: PatienceConfig, pos: source.Position): T = {
+    // SKIP-DOTTY-START 
+    def futureValue()(implicit config: PatienceConfig, pos: source.Position): T = {
+    // SKIP-DOTTY-END
+    //DOTTY-ONLY def futureValue()(implicit config: PatienceConfig, pos: source.Position): T = {  
       futureValueImpl(pos)(config)
     }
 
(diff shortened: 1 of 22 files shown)


jvm/core/src/main/scala/org/scalatest/concurrent/Futures.scala
```diff
@@ -472,7 +472,10 @@ trait Futures extends PatienceConfiguration {
      * @throws TestFailedException if the future is cancelled, expires, or is still not ready after
      *     the specified timeout has been exceeded
      */
-    def futureValue(implicit config: PatienceConfig, pos: source.Position): T = {
+    // SKIP-DOTTY-START 
+    def futureValue()(implicit config: PatienceConfig, pos: source.Position): T = {
```
bvenners, line 476: Why are empty parens added here? This is a breaking change to existing source code.

catap:
@cheeseng any chance to include scala3 for scala.js?

jvm/core/src/main/scala/org/scalatest/concurrent/Futures.scala
cheeseng: @bvenners Ah, sorry that was a mistake, the intention was to keep the Scala 2 version as it is, and add it only for Scala 3.

bvenners: Hmm, is it needed for Scala 3?

cheeseng: @bvenners Yes, if I remember correctly, that was needed to get our JavaFuturesSpec to build and green, in the following commit: 

https://github.com/scalatest/scalatest/pull/1966/commits/e32243bf9dd4dcdfd59bae7a29ccac209358b30b

The JavaFuturesSpec refused to compile, if I remember correctly it was confused about the implicit parameters.

Having said so, it was 3.0.0-M3, we can try again 3.0.0-RC1 I suppose.

cheeseng:
> @cheeseng any chance to include scala3 for scala.js?

The last time I tried there's an assertion error thrown from the scala-js compiler, so it does not seem like a straight forward thing.  We may try again with 3.0.0-RC1 though.

catap:
as far as I can see you're releasing support of 3.0.0-RC1 and Scala-Native 0.4.0 right now.

I guess it is worth to try to enable support Scala.js at Scala3.

Issue #1971, opened 2021-02-25
Can't run JUnit suites on 3.2.2


gabrielrussoc:
I have the following JUnit suite:

```
import org.junit.Assert;
import org.junit.Test;

public class ExampleSuite extends Assert {

    @Test
    public void success(){
    }

    @Test
    public void failure(){
	int x = 1/0;
    }
}
```

If I try to run it as `org.scalatest.tools.Runner.main(-j, ExampleSuite, -u, report, -oDF)`, following the [using the runner](https://www.scalatest.org/user_guide/using_the_runner) docs, it fails with:

```
*** RUN ABORTED *** (19 milliseconds)
  java.lang.ClassNotFoundException: org.scalatest.junit.JUnitWrapperSuite
  at java.base/java.net.URLClassLoader.findClass(URLClassLoader.java:435)
  at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:589)
  at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:522)
  at org.scalatest.tools.Runner$.doRunRunRunDaDoRunRun(Runner.scala:1208)
  at org.scalatest.tools.Runner$.$anonfun$runOptionallyWithPassFailReporter$24(Runner.scala:993)
  at org.scalatest.tools.Runner$.$anonfun$runOptionallyWithPassFailReporter$24$adapted(Runner.scala:971)
  at org.scalatest.tools.Runner$.withClassLoaderAndDispatchReporter(Runner.scala:1480)
  at org.scalatest.tools.Runner$.runOptionallyWithPassFailReporter(Runner.scala:971)
  at org.scalatest.tools.Runner$.main(Runner.scala:775)
  at org.scalatest.tools.Runner.main(Runner.scala)
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
  at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:64)
  at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  at java.base/java.lang.reflect.Method.invoke(Method.java:564)
  at org.codehaus.mojo.exec.ExecJavaMojo$1.run(ExecJavaMojo.java:297)
  at java.base/java.lang.Thread.run(Thread.java:832)
```
  
[...]

PR #1974, opened 2021-03-06
Release 3.2.6

cheeseng:
- Fixed class loading problem for JUnit and TestNG wrapper class in Runner as reported in https://github.com/scalatest/scalatest/issues/1971 .
- Changed to use scala-xml 2.0.0-M5 for Scala 3 build.
- Changed to use scala-xml 1.3.0 for Scala 2 build.


gabrielrussoc:
@cheeseng sorry to ping on this old thread, but is there any chance we can backport this to 3.1.x?

Reason being it's not possible to run JUnit suites on 3.1.x and the migration from 3.0 -> 3.2 is much easier if done in two steps: 3.0 -> 3.1 and 3.1 -> 3.2 in our case (we have some complicated inter-repo dependency)

PR #2011, opened 2021-04-22
[3.3] Safe/Pure Style Traits

cheeseng:
Please note that this branch is based on the following open PR: 

https://github.com/scalatest/scalatest/pull/2003

Please merge PR #2003 so that the diff here is clearer.

This PR brings back safe/pure style traits.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/enablers/InspectorAsserting.scala
+++ b/jvm/core/src/main/scala/org/scalatest/enablers/InspectorAsserting.scala
@@ -343,22 +343,23 @@ abstract class UnitInspectorAsserting {
  * Abstract class that in the future will hold an intermediate priority <code>InspectorAsserting</code> implicit, which will enable inspector expressions
  * that have result type <code>Expectation</code>, a more composable form of assertion that returns a result instead of throwing an exception when it fails.
  */
-/*abstract class ExpectationInspectorAsserting extends UnitInspectorAsserting {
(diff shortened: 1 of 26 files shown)


bvenners:
@cheeseng Actually the intention is to bring back FunSuite, not add a SafeFunSuite. We can now bring back plain old FunSuite as a style that takes Assertion as its result type, because it has been long enough since we deprecated the old FunSuite and moved folks over to AnyFunSuite. Can you do an update to this PR that changes the SafeX names to just X? Tnx.

cheeseng:
@bvenners Noted, I'll continue to work on this branch to take away the 'Safe' prefix.

cheeseng:
@bvenners I have dropped the 'Safe' prefix, except for SafeTestRegistration and FixtureSafeTestRegistration, do you think we should use 'Pure' for them?

bvenners:
> @bvenners I have dropped the 'Safe' prefix, except for SafeTestRegistration and FixtureSafeTestRegistration, do you think we should use 'Pure' for them?

@cheeseng These were deprecated in 3.1.0 with no replacement planned. I think we should remove them in 3.3.0. If all else looks good in this PR, I'll merge this, and make another ticket for removing the TestRegistration traits. For context, these traits were so people could do certain kinds of integrations without needing to pick a style. But they didn't get used much if at all, so to simplify the API we wanted to just drop them after a deprecation cycle.

PR #2041, opened 2021-06-18
Fixed a bunch of broken scaladoc links

kelemensanyi:
To improve developer experience while browsing the scaladoc on [scalatest.org](https://www.scalatest.org/) this PR aims to fix a lot of broken links.


kelemensanyi:
This is all I planned to add for this PR.

cheeseng:
@kelemensanyi This is helpful!  Thanks for helping, they looks good to me, I think what's still broken are those fixture/<style-trait>.html as they are no longer there, I'll recommend @bvenners to merge this first.

Cheers.

PR #2051, opened 2021-06-29
Flickers Tagging

cheeseng:
Added a Flicker tag and tag as many flicker as I could and make them to run only when SCALATEST_RUN_FLICKER_TESTS environment variable is set to true.

### DIFF
--- a/README.md
+++ b/README.md
@@ -40,6 +40,13 @@ This command will build and run the regular tests:
 
   `$ sbt test`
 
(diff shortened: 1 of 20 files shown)


SethTisue:
@bvenners would be awesome to see this merged

SethTisue:
I'm trying it out over at https://github.com/scala/community-build/pull/1459

PR #2070, opened 2021-10-23
Scala 3.1.0

cheeseng:
Updated to build with Scala 3.1.0.


cheeseng:
Let's update to Scala 3.1.0 only in 3.3.0.

PR #2079, opened 2021-12-11
Add missing semicolon to Tag.scala scaladoc

wojda:
An example from `org.scalatest.Tag` is missing a semicolon. This PR is adding that semicolon.


artimasites:
@cla-bot[bot] check

PR #2083, opened 2021-12-26
Scala 3: Tweak EitherValues and OptionValues Valuable Class Name

cheeseng:
Changed Valuable class in EitherValues to EitherValuable, and in OptionValues to OptionValuable, for Scala 3 build to avoid name crash error when both EitherValues and OptionValues are being used together.


cheeseng:
@bvenners Yes this will break binary compat for Scala 3 build only, the full change (including Scala 2) is being done in another PR for 3.3 release, current Scala 3 build for 3.2.10 will have the name crash problem described in the above ticket which imho is worse than the binary compat break.

PR #2086, opened 2022-01-03
Handle Non-Serializable Exception Payload

cheeseng:
Set non-serializable event's payload field to None and wrapped non-serializable throwable as NotSerializableWrapperException when failed to serialize them, and print out a warning message on both the client and server side so that user can follow-up to make the related throwable or payload type serializable, if they needs them to.  This approach avoid the non-serializable payload or throwable from crashing the test run.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/events/Event.scala
+++ b/jvm/core/src/main/scala/org/scalatest/events/Event.scala
@@ -281,6 +281,23 @@ sealed abstract class Event extends Ordered[Event] with Product with Serializabl
       }
     }
   }
+
+  private[events] def serializeRoundtrip(a: Any): Boolean = {
+    try {
+      val baos = new java.io.ByteArrayOutputStream
+      val oos = new java.io.ObjectOutputStream(baos)
+      oos.writeObject(a)
+      oos.flush()
+      val ois = new java.io.ObjectInputStream(new java.io.ByteArrayInputStream(baos.toByteArray))
+      ois.readObject
+      true
+    }
+    catch {
+      case _: Throwable => false
+    }
+  }
+
+  private[scalatest] def ensureSerializable(): Event
 }
 
 /**
@@ -395,6 +412,15 @@ final case class TestStarting (
     import EventJsonHelper._
     s"""{ "eventType": "TestStarting", "ordinal": ${ordinal.runStamp}, "suiteName": ${string(suiteName)}, "suiteId": ${string(suiteId)}, "suiteClassName": ${stringOption(suiteClassName)}, "testName": ${string(testName)}, "testText": ${string(testText)}, "formatter": ${formatterOption(formatter)}, "location": ${locationOption(location)}, "rerunner": ${stringOption(rerunner)}, "threadName": ${string(threadName)}, "timeStamp": ${timeStamp} }""".stripMargin
   }
+
+  private[scalatest] def ensureSerializable(): Event = 
+    payload match {
+      case Some(p) if !serializeRoundtrip(p) =>
+        println("Warning: Unable to serialize payload of type " + p.getClass().getName() + " for " + this.toString() + ", setting it to None.")
+        copy(payload = None)
+
+      case _ => this
+    }
 }
 
 /**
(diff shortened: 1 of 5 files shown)


jvm/core/src/main/scala/org/scalatest/events/Event.scala
```diff
@@ -281,6 +281,23 @@ sealed abstract class Event extends Ordered[Event] with Product with Serializabl
...
+      val oos = new java.io.ObjectOutputStream(baos)
+      oos.writeObject(a)
+      oos.flush()
+      val ois = new java.io.ObjectInputStream(new java.io.ByteArrayInputStream(baos.toByteArray))
+      ois.readObject
+      true
+    }
+    catch {
+      case _: Throwable => false
+    }
```
bvenners, line 297: @cheeseng We shouldn't catch Throwable, but something more specific. This will turn an OutOfMemoryError into false, for example. Any reason this isn't catching NotSerializableException?

jvm/core/src/main/scala/org/scalatest/tools/Framework.scala
```diff
@@ -886,8 +885,23 @@ import java.net.{ServerSocket, InetAddress}
...
           }
+
+          @annotation.tailrec
+          final def tryReact(): Unit = 
+            try {
+              react()  
+            }
+            catch {
+              case t: Throwable => 
+                // Restart server socket
```
bvenners, line 898: @cheeseng Here's another catch of Throwable. Should this be NotSerializableException?

jvm/core/src/main/scala/org/scalatest/tools/SocketReporter.scala
```diff
@@ -15,28 +15,50 @@
...
+        out.get.writeObject(event)
+        out.get.flush()
+      }
+      catch {
+        case e: java.io.NotSerializableException =>
+          refresh()
+          out.get.writeObject(event.ensureSerializable())
+
+        case e: Throwable => refresh()
+      }
```
bvenners, line 54: @cheeseng Here's another catch of Throwable. If you want to handle most exceptions uniformly, then I'd suggest Exception on Scala's NonFatal extractor.

jvm/core/src/main/scala/org/scalatest/events/Event.scala
```diff
@@ -502,6 +528,15 @@ final case class TestSucceeded (
...
+
+  private[scalatest] def ensureSerializable(): Event = 
+    payload match {
+      case Some(p) if !serializeRoundtrip(p) =>
+        println(Resources.unableToSerializePayload(p.getClass().getName(), this.toString()))
+        copy(payload = None)
+
+      case _ => this
+    }
 }
```
bvenners, line 540: @cheeseng Seems like quite a bit of duplication. Maybe factor out a method that takes params and does this match?

jvm/core/src/main/scala/org/scalatest/tools/Framework.scala
```diff
@@ -886,8 +885,23 @@ import java.net.{ServerSocket, InetAddress}
...
+                println(Resources.unableToReadSerializedEvent)
+                is.get.close()
+                socket.set(server.accept())
+                is.set(new SkeletonObjectInputStream(socket.get.getInputStream, getClass.getClassLoader))
+                tryReact()
+            }
```
bvenners, line 904: @cheeseng Should this loop forever in case it keeps getting an exception? At some point maybe it should exit?

cheeseng:
@bvenners I have pushed the adjustments, do you mind to review and consider again this PR for merging?

Thanks.

jvm/core/src/main/scala/org/scalatest/tools/SocketReporter.scala
```diff
@@ -15,28 +15,50 @@
...
+
+  def refresh(): Unit = {
+    try {
+      out.get.close()
+      socket.get.close()
+    }
+    catch {
+      case _: Throwable =>
+    }
+    socket.set(new Socket(host, port))
```
bvenners, line 38: @cheeseng I would us NonFatal here. We have one in ScalaTest, and Scala has one. I suppose we should use the ScalaTest one, which predated the Scala one, for internal consistency.

cheeseng:
@bvenners I have added the pattern guard according to your comment, do yo mind to review this again?

Thanks!

Issue #2069, opened 2021-10-22
Running org.scalatest.tools.Runner with an invalid Reporter shows exception but then hangs


jackkoenig:
I have a custom reporter that I provide to scalafmt using the CLI (no SBT or other build tool involved). This is a bit problematic because if I screw something up in CI, my CI just hangs. It would be ideal if the `Runner` would exit with a non-zero code rather than hanging.

cheeseng:
@jackkoenig Sorry for replying late on this, I am curious what is causing the hang, is it still possible to get the stacktrace of the last error when it hangs?

Cheers.

jackkoenig:
No worries and thanks for your response @cheeseng! I should have included it before, here's the stack trace:
```
An exception or error caused a run to abort. This may have been caused by a problematic custom reporter.
java.lang.IllegalArgumentException: Couldn't load a Reporter class: "my.custom.ReporterButMisspelt".
Problem arg: -r... my.custom.ReporterButMisspelt
        at org.scalatest.tools.ReporterFactory.getCustomReporter(ReporterFactory.scala:39)
        at org.scalatest.tools.ReporterFactory.createCustomReporter(ReporterFactory.scala:213)
        at org.scalatest.tools.ReporterFactory.getReporterFromConfiguration(ReporterFactory.scala:246)
        at org.scalatest.tools.ReporterFactory.$anonfun$createReportersFromConfigurations$1(ReporterFactory.scala:254)
        at scala.collection.TraversableLike.$anonfun$map$1(TraversableLike.scala:286)
        at scala.collection.Iterator.foreach(Iterator.scala:943)
        at scala.collection.Iterator.foreach$(Iterator.scala:943)
        at scala.collection.AbstractIterator.foreach(Iterator.scala:1431)
        at scala.collection.IterableLike.foreach(IterableLike.scala:74)
        at scala.collection.IterableLike.foreach$(IterableLike.scala:73)
        at org.scalatest.tools.ReporterConfigurations.foreach(ReporterConfiguration.scala:42)
        at scala.collection.TraversableLike.map(TraversableLike.scala:286)
        at scala.collection.TraversableLike.map$(TraversableLike.scala:279)
        at org.scalatest.tools.ReporterConfigurations.map(ReporterConfiguration.scala:42)
        at org.scalatest.tools.ReporterFactory.createReportersFromConfigurations(ReporterFactory.scala:253)
[...]

PR #2090, opened 2022-01-08
Fix Runner Hang Problem

cheeseng:
Fire RunAborted event when error in creating dispatch reporter, fixing hang problem when it happens.

https://github.com/scalatest/scalatest/issues/2069#issuecomment-1007611928

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/tools/Runner.scala
+++ b/jvm/core/src/main/scala/org/scalatest/tools/Runner.scala
@@ -1489,8 +1489,17 @@ object Runner {
         // getDispatchReporter may complete abruptly with an exception, if there is an problem trying to load
         // or instantiate a custom reporter class.
         case ex: Throwable => {
-          System.err.println(Resources.bigProblemsMaybeCustomReporter)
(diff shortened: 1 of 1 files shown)


jvm/core/src/main/scala/org/scalatest/tools/Runner.scala
```diff
@@ -1489,8 +1489,17 @@ object Runner {
         // getDispatchReporter may complete abruptly with an exception, if there is an problem trying to load
         // or instantiate a custom reporter class.
         case ex: Throwable => {
-          System.err.println(Resources.bigProblemsMaybeCustomReporter)
```
bvenners, line 1492: I think this is an improvement, for sure, but this is another catch of Throwable. Should we be using NonFatal here also?

cheeseng: @bvenners The Throwable was there before this PR, I think we did want to catch everything, as here we're going to fire RunAborted event to stop the run properly, if it is a fatal exception (like out of memory) that can cause the following code to not able to execute, I think it will be blowing up again here.

PR #2093, opened 2022-01-12
Scala XML 2.0.1

cheeseng:
Updated to scala-xml 2.0.1 for those that's possible.

### DIFF
--- a/project/DottyBuild.scala
+++ b/project/DottyBuild.scala
@@ -233,7 +233,7 @@ trait DottyBuild { this: BuildCommons =>
       initialCommands in console := """|import org.scalatest._
                                        |import org.scalactic._
                                        |import Matchers._""".stripMargin,
-      libraryDependencies += "org.scala-lang.modules" %%% "scala-xml" % "2.0.0", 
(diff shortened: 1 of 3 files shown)


bvenners:
@cheeseng I'm consistently getting this test failure with this PR:

```
[info] WaitersSpec:
[info] A Waiter 
[info] - should pass through even exceptions that would not normally cause a test to fail *** FAILED *** (71 milliseconds)
[info]   java.lang.NullPointerException:
[info]   at org.scalatest.concurrent.WaitersSpec.$anonfun$new$16(WaitersSpec.scala:130)
[info]   at scala.runtime.java8.JFunction0$mcV$sp.apply(JFunction0$mcV$sp.scala:18)
[info]   at org.scalatest.concurrent.Conductors$Conductor.$anonfun$threadNamed$1(Conductors.scala:545)
[info]   at org.scalatest.concurrent.Conductors$Conductor$TestThread.run(Conductors.scala:577)
[info]   ...
```

Can you see if you can reproduce that using current 3.2.x-new? Thanks.

cheeseng:
@bvenners I can't reproduce the problem, it is one of the common flickers as far as I can tell, may be you can try to exit sbt and try sbt testQuiick?

PR #2108, opened 2022-03-21
Adapt test following Scala 3 bytecode change

smarter:
After https://github.com/lampepfl/dotty/pull/14686, Scala 3 emits all
classes (even local classes) as public, just like Scala 2. But unlike Scala 2,
the local class TestSpec in XmlSocketReporterSpec will be emitted with a
zero-argument constructor (in Scala 2 its constructor takes an
XmlSocketReporterSpec outer parameter which is never used which seems like a
bug). This matters because Scalatest considers a test suite "rerunnable" if it
has a zero argument public constructor. So to get the test suite to pass, we
need to check for the presence of this constructor just like SuiteRerunner does.


cheeseng:
Closing as this is now included as part of https://github.com/scalatest/scalatest/pull/2110

cheeseng:
Closing as this is now included as part of https://github.com/scalatest/scalatest/pull/2110

PR #2114, opened 2022-04-04
Update scala-xml to 2.1.0

catap:


catap:
@cheeseng seems this one is already in master and can be closed, am I right?

PR #2117, opened 2022-04-06
Fix #2116: Use an object identity instead of UUID in TestSortingReporter

### LINKED ISSUE #2116: Support of Scala.js 1.10.0
Scala.js released 1.10.0 which contains broking changes inside `java.lang.UUID` which forces to use `java.security.SecureRandom`: https://www.scala-js.org/news/2022/04/04/announcing-scalajs-1.10.0/

`TestSortingReporter` is using `UUID.randomUUID` which can't be used in sjs since this release, it fails on compiling like:
```
[error] Referring to non-existent class java.security.SecureRandom
[error]   called from private java.util.UUID$.csprng$lzycompute()java.util.Random
[error]   called from private java.util.UUID$.csprng()java.util.Random
[error]   called from java.util.UUID$.randomUUID()java.util.UUID
[error]   called from static java.util.UUID.randomUUID()java.util.UUID
[error]   called from org.scalatest.tools.TestSortingReporter.distributingTest(java.lang.String)void
[error]   called from org.scalatest.AsyncSuperEngine.runTestImpl(org.scalatest.Suite,java.lang.String,org.scalatest.Args,boolean,boolean,scala.Function2,scala.concurrent.ExecutionContext)org.scalatest.Status
[error]   called from static org.scalatest.wordspec.AsyncWordSpecLike$class.runTest(org.scalatest.wordspec.AsyncWordSpecLike,java.lang.String,org.scalatest.Args)org.scalatest.Status
[error]   called from
...
```

I not sure how this should be fixed on the right way:
 - by adding dependency from fake SecureRandom;
 - moving away from UUID.

Impact? Huge. `AsyncWordSpec` and similar things are affected.

@cheeseng

sjrd:
This PR targets `main` because it's the default branch. But it should probably target all active branches of ScalaTest. If you tell me what branch is the best to target to achieve that, I'll be happy to switch.

---

The UUIDs are only used as an in-memory identifier for `Slot`s. It is more efficient and more reliable to use an object identity instead, for that purpose.

This removes the dependency on `UUID.randomUUID()`, which unblocks using ScalaTest with Scala.js >= 1.10.0. See
https://github.com/scala-js/scala-js/security/advisories/GHSA-j2f9-w8wh-9ww4
and
https://github.com/scala-js/scala-js/issues/4657
for the reason why `randomUUID()` fails to link by default in recent Scala.js.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/tools/TestSortingReporter.scala
+++ b/jvm/core/src/main/scala/org/scalatest/tools/TestSortingReporter.scala
@@ -35,8 +35,8 @@ private[scalatest] class TestSortingReporter(suiteId: String, dispatch: Reporter
   
   // Chee Seng: What's the UUID for?
   // Each test gets one slot, but other events such as an info from an after an also get a slot i think.
-  case class Slot(uuid: UUID, eventList: ListBuffer[Event], completed: Boolean, completedEvent: Boolean, ready: Boolean)
-  
+  private case class Slot(uuid: UUID, eventList: ListBuffer[Event], completed: Boolean, completedEvent: Boolean, ready: Boolean)
+
   private val waitingBuffer = new ListBuffer[Slot]()
   private val slotMap = new collection.mutable.HashMap[String, Slot]()  // testName -> Slot
   @volatile private var completedTestCount = 0 // Called within synchronized. Don't need volatile and it wouldn't work anyway.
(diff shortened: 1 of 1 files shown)


artimasites:
@cla-bot[bot] check

catap (approved):

cheeseng:
@sjrd main branch is fine, I can cherry-pick this fix into 3.2.x-new branch, I'll start test this.

Thanks!

cheeseng:
@sjrd The commits and changes should already be in the main branch and 3.2.x-new (released), not sure why this PR is still not marked as merged, I'll close and re-open if it will marked as merged automatically.

cheeseng:
Test reopen.

sjrd:
Yes, sure.

PR #2146, opened 2022-07-07
Scala 3 Open Class for Style Classes

cheeseng:
Use open class for concrete classes for style classes.


wasowski:
It does look good to me (I am glad I did not attempted myself, because I do not understand this preprocessing mechanism between DOTTY and Scala 2... - would've used more time on this than it looked up-front to require ...)

bvenners (approved):

PR #2151, opened 2022-07-21
3.2.13 Adjustments

cheeseng:
Adjustments for 3.2.13 release.


cheeseng:
Superseded by https://github.com/scalatest/scalatest/pull/2152 , we cannot update scalatest+testng to 3.2.12 yet because scalatest+testng 3.2.12.0 contains a bug that will be fixed by https://github.com/scalatest/scalatestplus-testng/pull/22 , since it is only used by our tests, let's only update that after the fix is released.

PR #2158, opened 2022-08-15
Scala 3 Infix

cheeseng:
Made all functions involved in matchers DSL infix functions.

### DIFF
--- a/dotty/matchers-core/src/main/scala/org/scalatest/matchers/dsl/BeWord.scala
+++ b/dotty/matchers-core/src/main/scala/org/scalatest/matchers/dsl/BeWord.scala
@@ -234,7 +234,7 @@ final class BeWord {
    * </p>
    */
   @deprecated("The deprecation period for the be === syntax has expired. Please use should equal, should ===, shouldEqual, should be, or shouldBe instead.")
-  infix def ===(right: Any)(implicit pos: source.Position): Matcher[Any] = {
+  def ===(right: Any)(implicit pos: source.Position): Matcher[Any] = {
     throw new NotAllowedException(FailureMessages.beTripleEqualsNotAllowed, pos)
   }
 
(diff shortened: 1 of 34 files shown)


dotty/matchers-core/src/main/scala/org/scalatest/matchers/dsl/BeWord.scala
```diff
@@ -90,7 +90,7 @@ final class BeWord {
    *                       ^
    * </pre>
    **/
-  def <[T : Ordering](right: T): Matcher[T] =
+  infix def <[T : Ordering](right: T): Matcher[T] =
     new Matcher[T] {
```
bvenners, line 94: I would remove infix from methods with symbolic names, because it is not required for symbols.

dotty/matchers-core/src/main/scala/org/scalatest/matchers/dsl/BeWord.scala
```diff
@@ -352,7 +352,7 @@ final class BeWord {
    *                      ^
    * </pre>
    **/
-  def apply[U](spread: Spread[U]): Matcher[U] =
+  infix def apply[U](spread: Spread[U]): Matcher[U] =
     new Matcher[U] {
```
bvenners, line 356: Hmm, I'm curious what syntax infix would support here?

cheeseng:
@bvenners Thanks for the comments!  I'll investigate and do the adjustments.

Cheers.

cheeseng:
@bvenners I have made the adjustments, do you mind to review again?

Thanks!

dotty/matchers-core/src/main/scala/org/scalatest/matchers/dsl/NotWord.scala
```diff
@@ -54,7 +54,7 @@ final class NotWord {
    *                     ^
    * </pre>
    **/
-  def apply[S](matcher: Matcher[S]): Matcher[S] =
+  infix def apply[S](matcher: Matcher[S]): Matcher[S] =
     new Matcher[S] {
```
bvenners, line 58: This infix should be removed.

cheeseng: @bvenners I have removed the related infix in the new commit that I just pushed.

Thanks for checking!

bvenners (approved):

PR #2161, opened 2022-08-21
Set sorting reporter as complete if test run failed unexpectedly

nimatrueway:
If test runner reporter is of "sorted" type, it buffers all events and eventually emits/prints them upon completion of the test. Now, in case test runner fails unexpectedly it does not flush buffered events until an internal timer kicks off. This won't work with SBT system as all loggers of a task are already discarded when a task is finished.

Sample
```scala
class BugBefore extends wordspec.AsyncWordSpec with BeforeAndAfter {

  before {
    fail("BeforeEach failure.")
  }

  "BeforeAndAfter" should {
    "print full stack trace if failed at 'before' stage." in {
      Future {
        succeed
      }
    }
  }
}
```

Before this PR
scalatest-before

After this PR
New Project

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/tools/Framework.scala
+++ b/jvm/core/src/main/scala/org/scalatest/tools/Framework.scala
@@ -206,7 +206,7 @@ class Framework extends SbtFramework {
    * @return <code>ScalaTest</code>
    */
   def name = "ScalaTest"
- 
+
   private val resultHolder = new SuiteResultHolder()
 
   /**
@@ -222,12 +222,12 @@ class Framework extends SbtFramework {
         def superclassName = "org.scalatest.Suite"
         def isModule = false
         def requireNoArgConstructor = true
-      }, 
+      },
       new AnnotatedFingerprint {
         def annotationName = "org.scalatest.WrapWith"
         def isModule = false
       })
-      
+
   private def runSuite(
     taskDefinition: TaskDef,
     rerunSuiteId: String,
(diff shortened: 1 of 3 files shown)


cheeseng:
@nimatrueway This PR looks good to me, thanks!

@bvenners Unfortunately this PR contains a lot of whitespace diff, in order to filter out whitespace diff lines you may apply the filter like this: 

Screenshot from 2022-08-22 22-55-05

nimatrueway:
@cheeseng thanks for reviewing it! I can remove all whitespace changes if it helps but I figured this formatting will happen eventually.

jvm/core/src/main/scala/org/scalatest/tools/Framework.scala
```diff
@@ -337,8 +337,9 @@ class Framework extends SbtFramework {
         }
       }
     }
-    catch {       
+    catch {
       case e: Throwable => {
+        suiteSortingReporter.completedTests(suite.suiteId)
```
bvenners, line 343: I wonder if we call this again after the timer expires. It sounds like it may not matter with running with sbt, but I wonder if that causes issues when running not with sbt.

nimatrueway: Hey Bill!

`completedTests` method invokes `fireReadyEvents` which in turn calls `cancelTimeoutTask` to cancel the subject **timer**. Whether **timeout** task is thread-safe is a more difficult question which I'm afraid I can't answer. However, that could happen regardless of this PR's change I believe. Namely, a test suite completes in the happy code path and simulatenously **timeout** task gets fired after 2 sec.

cheeseng: @bvenners @nimatrueway Framework.scala is called by sbt only, anyway, that also raises a question if we should do the similar when run with Runner, for Runner I think the similar spot is in SuiteRunner.scala: 

https://github.com/scalatest/scalatest/blob/main/jvm/core/src/main/scala/org/scalatest/tools/SuiteRunner.scala#L79

I am not sure how we can call completedTests there, we probably shouldn't.  Different from the sbt runner, the ScalaTest Runner will wait until the timeout and got the SuiteAborted eventually, I think.

bvenners (approved):

PR #2177, opened 2022-10-30
Differ Enhancement

cheeseng:
Enhanced GenSeqDiffer, GenSetDiffer and GenMapDiffer to support find diff elements according to limit size of prettifier when available.

### DIFF
--- a/jvm/scalactic/src/main/scala/org/scalactic/Differ.scala
+++ b/jvm/scalactic/src/main/scala/org/scalactic/Differ.scala
@@ -173,7 +173,7 @@ private[scalactic] class GenSetDiffer extends Differ {
         val limitedMissingInLeft: List[String] = if (missingInLeft.length > limit) missingInLeft.take(limit) :+ "..." else missingInLeft
 
         val shortName = Differ.simpleClassName(aSet)
-        println("####shortName: " + shortName)
+        println("####shortName: " + shortName + ", ")
         if (missingInLeft.isEmpty && missingInRight.isEmpty)
           PrettyPair(prettifier(a), prettifier(b), None)
         else {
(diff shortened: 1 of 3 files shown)


jvm/scalactic/src/main/scala/org/scalactic/Differ.scala
```diff
@@ -154,17 +160,27 @@ private[scalactic] class GenSetDiffer extends Differ {
...
+            case tp: TruncatingPrettifier => tp.sizeLimit.value
+            case _ => math.max(aSet.size, bSet.size)
+          }
+
+        val limitedMissingInRight: List[String] = if (missingInRight.length > limit) missingInRight.take(limit) :+ "..." else missingInRight
+        val limitedMissingInLeft: List[String] = if (missingInLeft.length > limit) missingInLeft.take(limit) :+ "..." else missingInLeft
 
         val shortName = Differ.simpleClassName(aSet)
+        println("####shortName: " + shortName + ", ")
         if (missingInLeft.isEmpty && missingInRight.isEmpty)
```
bvenners, line 177: Did you intend to remove this println?

jvm/scalactic/src/main/scala/org/scalactic/Differ.scala
```diff
@@ -179,43 +195,46 @@ private[scalactic] object GenSetDiffer extends GenSetDiffer
...
-            }.toSet ++
-            missingKeyInRight.flatMap { k =>
-              val leftValue = aMap(k)
-              Option(k.toString + ": " + leftValue + " -> ")
-            }.toSet
+        val limit = 
+          prettifier match {
+            case tp: TruncatingPrettifier => tp.sizeLimit.value
+            case _ => math.max(aMap.size, bMap.size)
+          }
```
bvenners, line 230: This match seems to be repeated 3 times. Perhaps factor out into a method?

cheeseng:
@bvenners Thanks for reviewing, I have made the adjustments can you see if it is fit now?

Thanks!

PR #2201, opened 2022-11-27
Canonical Lazy Shrink Batch 2

cheeseng:
Second batch of canonical lazy shrinks implementation.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
+++ b/jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
@@ -2134,8 +2134,22 @@ object Generator {
             (NextRoseTree(nonZeroFloat), Nil, rnd3)
         }
       }
-      private val nonZeroFloatCanonicals: LazyListOrStream[NonZeroFloat] = LazyListOrStream(NonZeroFloat(-3.0f), NonZeroFloat(3.0f), NonZeroFloat(-2.0f), NonZeroFloat(2.0f), NonZeroFloat(-1.0f), NonZeroFloat(1.0f))
-      override def canonicals(rnd: Randomizer): (LazyListOrStream[RoseTree[NonZeroFloat]], Randomizer) = (nonZeroFloatCanonicals.map(Rose(_)), rnd)
+      override def canonicals(rnd: Randomizer): (LazyListOrStream[RoseTree[NonZeroFloat]], Randomizer) = {
+        case class CanonicalRoseTree(value: NonZeroFloat) extends RoseTree[NonZeroFloat] {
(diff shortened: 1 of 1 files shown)


jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
```diff
@@ -2216,9 +2258,23 @@ object Generator {
             (NextRoseTree(nonZeroInt), Nil, rnd2)
         }
       }
+      override def canonicals(rnd: Randomizer): (LazyListOrStream[RoseTree[NonZeroInt]], Randomizer) = {
+        case class CanonicalRoseTree(value: NonZeroInt) extends RoseTree[NonZeroInt] {
+          def shrinks(rndPassedToShrinks: Randomizer): (LazyListOrStream[RoseTree[NonZeroInt]], Randomizer) = {
+            def resLazyList(theValue: NonZeroInt): LazyListOrStream[RoseTree[NonZeroInt]] = {
+              if (theValue.value == 1 || theValue.value == -1) LazyListOrStream.empty
+              else {
```
bvenners, line 2266: @cheeseng I think if we check for both 1 and -1, we'll end up stopping at -1 every time, and we want to go to 1.

jvm/core/src/main/scala/org/scalatest/prop/Generator.scala
```diff
@@ -2253,8 +2309,22 @@ object Generator {
...
         }
       }
-      private val nonZeroLongCanonicals = LazyListOrStream(NonZeroLong(-3), NonZeroLong(3), NonZeroLong(-2), NonZeroLong(2), NonZeroLong(-1), NonZeroLong(1))
-      override def canonicals(rnd: Randomizer): (LazyListOrStream[RoseTree[NonZeroLong]], Randomizer) = (nonZeroLongCanonicals.map(Rose(_)), rnd)
+      override def canonicals(rnd: Randomizer): (LazyListOrStream[RoseTree[NonZeroLong]], Randomizer) = {
+        case class CanonicalRoseTree(value: NonZeroLong) extends RoseTree[NonZeroLong] {
+          def shrinks(rndPassedToShrinks: Randomizer): (LazyListOrStream[RoseTree[NonZeroLong]], Randomizer) = {
+            def resLazyList(theValue: NonZeroLong): LazyListOrStream[RoseTree[NonZeroLong]] = {
+              if (theValue.value == 1L || theValue.value == -1L) LazyListOrStream.empty
+              else {
```
bvenners, line 2317: @cheeseng Same here. Probably just check for 1L, but make sure we get there.

PR #2221, opened 2023-01-11
Custom Differ Support

cheeseng:
Added overridable differ to Prettifier trait, and added functions to create prettifier instance with supplied Differ.

### DIFF
--- a/jvm/scalactic-test/src/test/scala/org/scalactic/PrettifierSpec.scala
+++ b/jvm/scalactic-test/src/test/scala/org/scalactic/PrettifierSpec.scala
@@ -186,6 +186,17 @@ class PrettifierSpec extends funspec.AnyFunSpec with matchers.should.Matchers {
     it("should pretty print nested Many(String)") {
       Prettifier.basic(Many(Many("1", "2", "3"), Many("7", "8", "9"))) should be ("Many(Many(1, 2, 3), Many(7, 8, 9))")
     }
+    it("should support custom Differ") {
+      val d = 
+        new Differ {
+          def difference(a: Any, b: Any, prettifier: Prettifier): PrettyPair = {
+            PrettyPair(a.toString, b.toString, None)
+          }
+        }
+      val p = Prettifier(Prettifier.basic, d)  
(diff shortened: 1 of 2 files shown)


charpov:
How does one use a custom `Differ`? It looks like the `Differ` type is not public...

cheeseng:
@charpov I think you are right, @charpov at the moment perhaps the only way is to have your custom Differ use org.scalactic package.

@bvenners Shall we make Differ public in next 3.2 release?

charpov:
I don't know if this is the right place for this discussion (given the "merged" nature of the issue), but I still think the design could be improved.

I can now use a truncated prettifier do deal with large collections:

```
given defaultPrettifier: Prettifier = Prettifier.truncateAt(SizeLimit(32))
```

Then, in a test suite that uses large lists and large random trees, I further customize the prettifier:

```
given Prettifier = o =>
   o.asMatchable match
      case _: Tree[?] | _: Forest[?] => o.toString.short(512)
      case _                         => defaultPrettifier(o)
```

Now, I noticed that while assertions of the form `assertResult(l2)(l1)` don't, `assert(l1 == l2)` generates an "analysis". On large lists, this analysis still produces undesirable strings. It is trying to limit things, but uses a test:

```
case tp: TruncatingPrettifier => ...
```
and my prettifier is *not* a `TruncatingPrettifier` anymore.

It seems I can turn off this analysis with:

```
given Prettifier with
   def apply(o: Any): String = o.asMatchable match
      case _: Tree[?] | _: Forest[?] => o.toString.short(512)
      case _                         => defaultPrettifier(o)

   override def apply(left: Any, right: Any): PrettyPair =
      PrettyPair(apply(left), apply(right), None)
```

but that's a bit radical.

So, my questions:

- Is there a way to turn "analysis" on and off? The fact that `assert` has it but `assertResult` doesn't is confusing.
- Would it make sense for the `Prettifier` type to implement a limit (defaulting to `None`) and use it in `Differ` instead of checking that the type is `TruncatingPrettifier`? I could then specify a limit in my custom prettifier and rely on the existing behavior of the `Differ.difference` methods.

PR #2222, opened 2023-01-17
Recognize -S as a valid argument

inmost-light:
Ran into this problem while trying to reproduce a test failure from a property-based test.


artimasites:
@cla-bot[bot] check

cheeseng:
@inmost-light Thanks for the PR, I forgot to add back in checkArgsForValidity when I resurrected -S in this PR: 

https://github.com/scalatest/scalatest/pull/2104

It is being called by Runner, but not the SBT integration that I tested with.  Thanks for catching it.  :)

lionralfs:
I'm running into the same issue. Has this change been released yet?

cheeseng:
@lionralfs Unfortunately not yet, though the PR is merged.  It should be included in the upcoming 3.2.16.

PR #2227, opened 2023-02-21
maybe this is the correct file?

rebekah:
When I ran the RefSpec example from https://www.scalatest.org/user_guide/selecting_a_style, using scala 2.13.8, sbt 1.7.1, and scalatest 3.2.15... I got this error "procedure syntax is deprecated: instead, add : Unit = to explicitly declare..." I was able to fix it by adding the equals sign to the method definition. I am not sure if this is the file that needs to be updated. I just took a guess and went with it.


artimasites:
@cla-bot[bot] check

bvenners (approved):

PR #2228, opened 2023-02-23
NoArg Deprecation

cheeseng:
- Deprecated NoArg as scala.DelayedInit is dropped from Scala 3 already, 
- Updated scaladoc about the scala.DelayedInit being dropped in Scala 3.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/fixture/NoArg.scala
+++ b/jvm/core/src/main/scala/org/scalatest/fixture/NoArg.scala
@@ -135,8 +135,8 @@ package org.scalatest.fixture
  * <p>
  * Note: As of Scala 2.11, <code>DelayedInit</code> (which is used by <code>NoArg</code>) has been deprecated, to indicate it is buggy and should be avoided
  * if possible. Those in charge of the Scala compiler and standard library have promised that <code>DelayedInit</code> will not be removed from Scala
- * unless an alternate way to achieve the same goal is provided. Thus it <em>should</em> be safe to use <code>NoArg</code>, but if you'd rather
- * not you can achieve the same effect with a bit more boilerplate by extending (<code>() =&gt; Unit</code>) instead of <code>NoArg</code> and placing
+ * unless an alternate way to achieve the same goal is provided. Unfortunately, <code>DelayedInit</code> was officially dropped in Scala 3 (https://docs.scala-lang.org/scala3/reference/dropped-features/delayed-init.html). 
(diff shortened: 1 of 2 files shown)


mkurz (approved):

mkurz:
@cheeseng please wait with merging, I will come back to this soon.

bvenners (approved):

cheeseng:
> @cheeseng please wait with merging, I will come back to this soon.

Hi @mkurz , just checking if you have a better idea for this?

mkurz:
@cheeseng Sorry for the delay. Please don't remove `NoArg` from the Scala 3 build (even though it is doing nothing). I think the deprecation is enough to make people aware.

I want to keep it because in scalatestplus-play I want to offer people a smoother migration instead of having compiler errors where users maybe don't know what to do.
So if someone moves from Scala 2 to 3 (by simply changing the `scalaVersion` in `build.sbt`) what they will see in scalatestplus-play is following error (instead of compile error if you would remove `NoArg`):

```sh
[info] MixedFixtureSpec:
[info] org.scalatestplus.play.MixedFixtureSpec *** ABORTED ***
[info]   scala.NotImplementedError was thrown inside "The App function" must, construction cannot continue: "
[info]   
[info]   For Scala 3 you need to wrap the body of App in an `override def running() = ...` method:
[info]   
[info]   // Old:
[info]   new App() {
[info]     <code>
[info]   }
[info]   
[info]   // New:
[info]   new App() {
[info]     override def running() = {
[info]       <code>
[info]     }
[info]   }
[info]   
[info]   " (MixedFixtureSpec.scala:66)
```

I implemented that hint here: https://github.com/playframework/scalatestplus-play/pull/320/commits/36fdb5eb755a75096675084986f2b95f79789544
Ands wrapped the test methods here: https://github.com/playframework/scalatestplus-play/pull/320/commits/60ddc34462ac730b1b4193bc05961f30623cdf0e

Therefore I am asking you to not remove `NoArg` in the Scala 3 build but keep it wit the deprecation warning.

mkurz:
IMHO it should be fine to keep it. It was here all the time in the Scala 3 build and no one was bothered by it :wink:

bvenners:
@cheeseng Yes it should be fine to keep `NoArg` in the Scala 3 build, at least for a good long time. We used to keep deprecated things in the library for at least two years. With Scala 2 to 3, because people will take longer to upgrade, we should probably keep it even longer.

cheeseng:
@bvenners @mkurz fyi I have undo the NoArg removal from Scala 3 build.

Thanks!

mkurz (approved):

bvenners (approved):

Issue #1445, opened 2018-12-03
SuiteAborted must not assume that message is non-null


danile42:
When a org.scalatest.events.SuiteAborted event is created due to an exception that is thrown during the test, it cannot be assumed that the message coming from this exception is not null, because null is a valid value as documented in the JDK docs: "the detail message string of this Throwable instance (which may be null)".

Currently, ScalaTest (3.0.5) fails in this case with a NullArgumentException, the original exception becomes invisible.

Maybe a synthetic message has to be created in this case instead.

shawjef3:
I created #2236 to fix this. I verified it fixes the code that created the above stack trace.

bvenners:
Fixed by https://github.com/scalatest/scalatest/pull/2236.

PR #2236, opened 2023-04-14
fix "SuiteAborted.this.message was null"

shawjef3:
This is for #1445 .


cheeseng:
@bvenners LGTM, I think we should include this in the upcoming 3.2.16.

shawjef3:
@travisbrown are there other branches that need the same patch?

cheeseng:
@shawjef3 No worry, we'll pull changes from 3.2.x-new after this is merged into 3.2.x-new.

bvenners (approved):

PR #2239, opened 2023-05-29
Fix: Correction of the path of exceptions thrown in certain comments to correct the DRI problem for scaladoc.

Dedelweiss:
The aim of this PR is to correct link problems in certain comments in order to avoid a Scaladoc DRI problems. 
Indeed, the Scaladoc can't find out the paths of classes called in `throws` comments, so it needs the full path.
The examples below are taken from the Scaladoc generation of an external project using Scalatest.

## Before
Screenshot 2023-05-30 at 11 37 35

## After
Screenshot 2023-05-30 at 11 27 08

### DIFF
--- a/dotty/core/src/main/scala/org/scalatest/Assertions.scala
+++ b/dotty/core/src/main/scala/org/scalatest/Assertions.scala
@@ -467,7 +467,7 @@ trait Assertions extends TripleEquals  {
    * </p>
    *
    * @param condition the boolean condition to assert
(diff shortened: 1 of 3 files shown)


artimasites:
@cla-bot[bot] check

Dedelweiss:
IE: I'm currently pushing on the main one because I don't know where I can push it, but I can fix it no problem. And if anyone can run the workflow test, please. Thank you very much.

bvenners (approved):

PR #2241, opened 2023-06-03
Scala 3.3.0 for Main Branch

cheeseng:
Upgraded to use the latest Scala 3.3.0.


cheeseng:
Seems like scala-native does not have a version that works with scala 3.3.0 yet, but hopefully it will be soon.

bvenners (approved):

PR #2244, opened 2023-06-14
Make Expectations Public

cheeseng:
- Made Expectations Public.
- Added scaladoc to Expectations.


bvenners (approved):
@cheeseng I was thinking we might make Expectations a module, so that there is one fewer choice by default. Think about it and let's talk about it tomorrow.

cheeseng:
@bvenners I think it is a good idea, do you think we should include Fact in the new module also?

PR #2249, opened 2023-06-19
Rose Tree Shrink

cheeseng:
Shrinking with rose tree.


jvm/core/src/main/scala/org/scalatest/prop/RoseTree.scala
```diff
@@ -0,0 +1,208 @@
...
+  val value: T
+
+  // Compute the shrinks list on demand using this RoseTree's value.
+  // This will be called only when a property fails, and just once, and it
+  // won't take long, so no need to make this a lazy val.
+  def shrinks: LazyListOrStream[RoseTree[T]]
+
+  private lazy val maximumIterationCount = 1000000
+
+  def depthFirstShrinks[E](fun: T => (Boolean, Option[E])): (LazyListOrStream[RoseTree[T]], Option[E]) = {
```
bvenners, line 33: Let's call this shrinkSearch, because it really is neither depth first nor breadth first.

bvenners (approved):

PR #2284, opened 2023-09-24
Conductors Adjustment

cheeseng:
Try to catch UnsupportedOperationException when calling Thread's stop, and try to call method interrupt when that happens as next attempt to stop the running thread.  This unblocks the ConductorSuite when runs using JDK 21.


bvenners (approved):
The Java folks are probably trying to tell us to not call stop anymore, but let's keep this working. Conductor was always iffy, because it is guaranteed to work according to the JMM.

PR #2290, opened 2023-09-29
ScalaTest-App for Scala 3

cheeseng:
Added scalatest-app build for scala 3.


cheeseng:
@danghuybk It will work for Scala 3.3 as Scala 3 is forward compatible, but it won't work for Scala 3.0.x.

bvenners (approved):

PR #2291, opened 2023-10-02
By-Name Clue

cheeseng:
Changed clue argument in multiple places as by-name.


bvenners:
I'm going to close this one. Those clue params have been non-by-name for 15 years, so I think it isn't a major need. The workaround is to pass an object whose toString method causes the computation upon the first invocation.

PR #2293, opened 2023-10-05
Drop Chosen Style Traits

cheeseng:
- Drop the deprecated chosen style traits support.
- Removed deprecated styleName from Suite.

Currently when running tests for Scala 3, there's a NoSuchMethodException thrown when initializing JUnit3Suite, which should be due to the styleName removal from Suite.  I think after this is approved and merged into main, we can make a SNAP5 release, and a 3.3.0.0-SNAP5 release also for scalatest+junit (and junit5?) to compile against scalatest 3.3.0-SNAP5 that has styleName drop, and we update our main to use scalatest+junit 3.3.0.0-SNAP5.


bvenners (approved):
Wow, lots of changes to remove this feature.

Issue #2300, opened 2023-10-20
PrivateMethodTester does not correctly honor type bounds for method arguments


nickmoorman:
I recently ran into a situation where `PrivateMethod`/`PrivateMethodTester` does not seem to correctly honor the type bounds defined for arguments to the method being tested.  I put together the code below to illustrate a simplified example of what I'm seeing.

Here are some definitions that mimic the structure of the code used in my application:
```scala
// Abstract class and marker traits to set up test conditions
abstract class Thing
trait HasPropertyA
trait HasPropertyB

// Concrete implementations for tests
class ThingWithPropertyA extends Thing with HasPropertyA
class ThingWithPropertiesAAndB extends Thing with HasPropertyA with HasPropertyB

// Object containing methods to test
object TestClass {
  def doSomething[T <: Thing with HasPropertyA with HasPropertyB](
    thing: T
  ): String = {
    s"thing: $thing"
  }

  private def doSomethingPrivately[T <: Thing with HasPropertyA with HasPropertyB](
    thing: T
  ): String = doSomething(thing)
}
```

This test code (with comments showing the results when run) demonstrates the problem:
```scala
import org.scalatest.PrivateMethodTester
import org.scalatest.freespec.AnyFreeSpec
import org.scalatest.matchers.should.Matchers

class PrivateMethodTesterSpec extends AnyFreeSpec with Matchers with PrivateMethodTester {
  val thingA = new ThingWithPropertyA
  val thingAB = new ThingWithPropertiesAAndB

  // PASS
  "doSomething() should work for correctly typed Things" in {
    TestClass.doSomething(thingAB) shouldEqual s"thing: $thingAB"
  }

  // PASS
  "calling doSomething() with incorrectly typed Things should not type check" in {
    "TestClass.doSomething(thingA)" shouldNot typeCheck
  }

  val doSomethingPrivately = PrivateMethod[String](Symbol("doSomethingPrivately"))

  // PASS
  "doSomethingPrivately() should work for correctly typed Things" in {
    TestClass.invokePrivate(doSomethingPrivately(thingAB)) shouldEqual s"thing: $thingAB"
  }

  // FAIL
  "calling doSomethingPrivately() with incorrectly typed Things should not type check" in {
    "TestClass.invokePrivate(doSomethingPrivately(thingA))" shouldNot typeCheck
  }

  // PASS
  "calling doSomethingPrivately() with incorrectly typed Things should not type check but works anyway" in {
[...]

PR #2301, opened 2023-10-23
Safe PrivateMethodTester

cheeseng:
Safer PrivateMethodTester, with typed arguments vs the existing one that's use only Any.

Related ticket that complained about the problem: 

https://github.com/scalatest/scalatest/issues/2300


jvm/core/src/main/scala/org/scalatest/PrivateMethodTester.scala
```diff
@@ -102,10 +102,919 @@ trait PrivateMethodTester {
      * The type parameter, <code>T</code>, is the return type of the private method.
      *
      * @param args zero to many arguments to pass to the private method when invoked
-     * @return an <code>Invocation</code> object that can be passed to <code>invokePrivate</code> to invoke
-     * the private method
+     * @return an <code>Invocation0</code> object that can be passed to <code>invokePrivate</code> to invoke the private method
      */
-    def apply(args: Any*) = new Invocation[T](methodName, args: _*)
+    def apply() = new Invocation0[T](methodName)
+  }
```
bvenners, line 108: I think we should leave PrivateMethod alone so that no code breaks. We can decide to deprecate it, but that would give folks time to change to PrivateMethodN. Or we could not deprecate it and just let people choose. So please take PrivateMethod and plain old Invocation back to the way it was, then probably add a PrivateMethod0 trait that would use the Invocation0 object.

bvenners (approved):

PR #2305, opened 2024-01-01
Add method `checkpoint` to trait Checkpoints for easy use of `Checkpoint`

ndy2:
I added a method `checkpoint` to trait Checkpoints for easy use of `Checkpoint`.

I have some experience of other test frameworks of JVM familiy like Junit/assertj & kotest.
I thought that create Checkpoint & report it is quite verbose and tedious. 

Thus, I introduce a checkpoint!
I found that like the trait `Checkpoint` is right place to place this method like other dsl helper traits in scalatest. (`Inside`, `Matchers` ...)

Full Usage Example : 
```
class FourBasicOperationSpec extends AnyFlatSpec with Matchers with Checkpoints {

  it("four basic operation") {
    checkpoint { cp =>
      cp { 1 + 1 should equal(2) }
      cp { 2 + 2 should equal(4) }
    }

    checkpoint { cp =>
      cp { 1 - 1 should equal(0) }
      cp { 2 - 2 should equal(0) }
    }

   ...
  }
}
```

Inspired by 
- assertj - soft assertions :https://assertj.github.io/doc/#assertj-core-assertSoftly-assertions
- kotest - soft-assertions : https://kotest.io/docs/assertions/soft-assertions.html


artimasites:
@cla-bot[bot] check

ndy2:
Hello, @cheeseng !
I have no idea why the compilation fail in scala 2.13,2.11, 3, js even the unit tests are all passed.
Can you give me some suggestion or comment on this PR?

cheeseng:
@ndy2 Sorry for the late response, I'll take a look today and what's going on.

Cheers.

cheeseng:
@ndy2 I managed to fix the compile errors: 

https://github.com/cheeseng/scalatest/commit/4d2fcc5ed88293a807698eebe2232360a2d817e7

But the unit tests are failing because of wrong line number check, I'll try to look into it further when possible.

cheeseng:
@ndy2 I managed to fix the problem, I submitted the following PR to your fork: 

https://github.com/ndy2/scalatest/pull/1

Hopefully after merging it should get your PR pass the CI here.

Cheers.

ndy2:
I merged the PR in my fork.
Can I get approval for check the CI pass?

bvenners (commented):
This seems like a nice addition. But I would probably call the method withCheckpoint rather than checkpoint. And the signature should be:

```
def withCheckpoint(f: Checkpoint => Assertion)(implicit pos: source.Position): Assertion
```

Instead of:

```
def checkpoint(f: Checkpoint => Unit)(implicit pos: source.Position): Unit
```

ndy2:
@bvenners Hi. 

The suggested signature looks good. I agree that `withCheckpoint` is a better method name.

But how can we use `Assertion` as the Return type of both parameter `f` and `withCheckPoint` ? 
Do you mean `org.scalatest.compatible.Assertion` in `Assertion` above?

I found that `org.scalatest.compatible.Assertion` (call `Assertion`) do nothing and is a simple marker interface, but to apply suggested method with below usage, it requries to update the signature of `Checkpoints.apply` also (to return `Assertion`).

below usage
```scala
    withCheckpoint { cp =>
       cp { x should be < 0 }
       cp { y should be > 9 }
    }
```

or to write as below which is quite not awesome.

below
```scala
    withCheckpoint { cp =>
       cp { x should be < 0 }
       cp { y should be > 9 }
       Succeeded
    }
```

Can you explain what do you want to acheive by

bvenners:
> Can you explain what do you want to achieve by use `Assertion` instead of `Unit` in the signature, and how can handle above situation?

You're right. The problem is that Checkpoint's apply method returns Unit. It should probably return Assertion, but we did not change that when we added Assertion. Please leave it as Unit in your PR and we can use that for now. Perhaps later we can think about migrating to Assertion.

bvenners (approved):

Issue #2314, opened 2024-03-02
Execute tasks on SerialExecutionContext outside of a test


LeifW:
When using AsyncTestSuite (which provided an implicit ExectionContext with SerialExecutionContext), it may sometimes be necessary to run cleanup code after all the tests have executed - e.g. in `afterAll`. However, if that cleanup code involves dispatching to the `ExecutionContext`, those won't get run (and thus any `Await`s will time out).

`SerialExecutionContext` is private, so I can't invoke the `runNow` method directly on there. The way I've hacked around this for now is by reflectively calling `runNow` after the Futures have been dispatched in `afterAll`:
```scala
val runNow = PrivateMethod[Unit](Symbol("runNow"))
executionContext invokePrivate runNow(Future.successful(Succeeded))
```

Perhaps exposing something like that to be more conveniently called in `afterAll` would be nice? (Just making `SerialExecutionContext` not private would be an improvement).

I get that this is intended for use during tests, but ExecutionContexts aren't always passed granularly enough for me to use one during test code and another during cleanup / shutdown code.

It seems `ExecutionContext.parasitic` also accomplishes the goal of keeping things single-threaded, and it doesn't require me to "run" the tasks submitted to it. Is there any advantage to using `SerialExecutionContext` over `ExecutionContext.parasitic`?

cheeseng:
@LeifW `SerialExecutionContext` is just a simple form of default execution context, which was there before ExecutionContext.parasitic was introduced, and we still need to support older Scala version 2.10/2.11.  I think you may try to override executionContext with ExecutionContext.parasitic or your custom SerialExecutionContext by just copying the original code and modify.

bvenners:
@LeifW As Chee Seng mentioned, you can just override the implicit execution context with another one that serves your needs. I would try ExecutionContext.parasitic, which to my knowledge did not exist yet when we created the SerialExecutionContext, and see if that works for you.

PR #2316, opened 2024-03-04
Expose a method to run any queued-up tasks on the SerialExecutionContext

LeifW:
For calling from outside of a test when there are pending things dispatched to there - e.g. in `afterAll()` during cleanup.

`SerialExecutionContext` is private - and so this run method is inaccessible from outside of ScalaTest.

To address https://github.com/scalatest/scalatest/issues/2314


artimasites:
@cla-bot[bot] check

bvenners:
I'm going to close this one because SerialExecutionContext is just a default. The user can override the executionContext def with a different ExecutionContext, in which case serialExecutionContext is not used at all and the exposed executeSerialECTasks method would be a no-op. Better is for the user to override executionContext with an execution context that allows them to do what they need. For example, @LeifW suggested ExecutionContext.parasitic might solve his particular problem with needing to call runNow from afterAll.

PR #2318, opened 2024-03-13
Support scala-natve 0.5.0

catap:
It has a few changes in build plugin, which requires to cleanup a bit build script.

And one deprectated class was removed.

Tested as:
```
export SCALANATIVE_VERSION=0.5.0-RC2

sbt clean

sbt ++2.12.17 "project scalacticNative" clean publishLocal
sbt ++2.13.10 "project scalacticNative" clean publishLocal
sbt "project scalacticDottyNative" clean publishLocal

sbt ++2.12.17 "project scalatestNative" clean publishLocal
sbt ++2.13.10 "project scalatestNative" clean publishLocal
sbt "project scalatestDottyNative" clean publishLocal

sbt ++2.12.17 "project scalatestAppNative" clean publishLocal
sbt ++2.13.10 "project scalatestAppNative" clean publishLocal
sbt "project scalatestAppDottyNative" clean publishLocal
```

PR #2322, opened 2024-04-18
Scala Native 0.5 Build

cheeseng:
This PR is based on https://github.com/scalatest/scalatest/pull/2318 , just minor tweaked for publishing 3.2.18 build for scala-native 0.5.

### DIFF
--- a/project/DottyBuild.scala
+++ b/project/DottyBuild.scala
@@ -19,9 +19,13 @@ trait DottyBuild { this: BuildCommons =>
 
   // List of available night build at https://repo1.maven.org/maven2/ch/epfl/lamp/dotty-compiler_0.27/
   // lazy val dottyVersion = dottyLatestNightlyBuild.get
-  lazy val dottyVersion = System.getProperty("scalatest.dottyVersion", "3.3.3")
+  lazy val dottyVersion = System.getProperty("scalatest.dottyVersion", "3.1.3")
   lazy val dottySettings = List(
     scalaVersion := dottyVersion,
+    scalacOptions ++= List("-language:implicitConversions", "-noindent", "-Xprint-suspension")
+  )
+  lazy val nativeDottySettings = List(
+    scalaVersion := "3.3.3",
     scalacOptions ++= List("-language:implicitConversions", "-noindent", "-Xprint-suspension")
   )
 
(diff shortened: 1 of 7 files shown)


catap:
@cheeseng this introduces update of dependencies and to be on safe side I suggest to release it as 3.2.19.

cheeseng:
> @cheeseng this introduces update of dependencies and to be on safe side I suggest to release it as 3.2.19.

@catap I think it should be fine, there's a scala-native 0.4 already published previously for 3.2.18 and it still can be used, or user may choose to use the 0.5 build (staged) if want to.  Let's see if @bvenners likes it, starting 3.2.19 we should just have build for 0.5 though.  :)

catap:
On Thu, 18 Apr 2024 17:10:30 +0200,
Chua Chee Seng wrote:
> 
> > @cheeseng this introduces update of dependencies and to be on safe side I suggest to release it as 3.2.19.
> 
> @catap I think it should be fine, there's a scala-native 0.4 already published previously for 3.2.18 and it still can be used, or user may choose to use the 0.5 build (staged) if want to.  Let's see if @bvenners likes it, starting 3.2.19 we should just have build for 0.5 though.  :)
> 

Keep in mind that here I've updated scala-xml for example, and to have
3.2.18 which depends on different version of scala-xml may lead to issues.

-- 
wbr, Kirill

catap:
@cheeseng unfortently I was forced to upgrade scala-js to match version of sjs inside dependencies like scala-xml.

project/DottyBuild.scala
```diff
@@ -289,7 +289,7 @@ trait DottyBuild { this: BuildCommons =>
       initialCommands in console := """|import org.scalatest._
                                        |import org.scalactic._
                                        |import Matchers._""".stripMargin,
-      libraryDependencies += "org.scala-lang.modules" %%% "scala-xml" % "2.3.0",
+      libraryDependencies += "org.scala-lang.modules" %%% "scala-xml" % "2.1.0",
```
catap, line 292: ah, this is dedicated project for Scala3Js. Got it.

bvenners (approved):

cheeseng:
@catap Fyi Artifacts for scala-native 0.5 are up.

PR #2366, opened 2025-05-02
add `ResourceManagerFixture` trait, for resource management with `scala.util.Using.Manager`

mberndt123:
Hey there,

As a long-time ScalaTest user, there are some issues with its support for resource management using `BeforeAndAfterAll` and `BeforeAndAfterEach` that have been bothering me for a long time.
 ## Motivation
These are:
 - the need for `var` and `null`. When you want to use `BeforeAndAfterAll` to initialize a shared (between tests in a suite) resource, you basically have to use a `var` initialized to `null` that you assign to in `beforeAll` and destroy in `afterAll`. Both `null` and `var` are widely considered something to avoid if possible
 - Fixture traits just don't work as a unit of composition. For example, you might have a trait that provides a database container fixture, and another trait that provides a connection pool fixture. You can't use the two together because there is only one `FixtureParam` type. Some libraries like testcontainers-scala-scalatest work around this by defining their own mechanisms for composition, and of course none of these are interoperable
 - resource management with `BeforeAndAfter*` is awkward and brittle. When overriding `before*` or `after*` you have to remember calling `super.before*` and `super.after*`, and you need to do it in the right order, which it isn't even clear what that is. Order of evaluation depends on the order that you mix in the traits, which is subtle and easy to get wrong. And when one of multiple stacked traits fails in its `before*` method, how do you make sure that previously allocated resources are cleaned up?

For these reasons I think another approach is needed.

## Design

`ResourceManagerFixture` is a trait that can be mixed into any `FixtureTestSuite`.

It provides a [`Using.Manager`](https://www.scala-lang.org/api/2.13.16/scala/util/Using$.html) to every test as a fixture parameter, and the protected `suiteScoped` method will give you access to a `Using.Manager` that will be cleaned up after all tests in the suite have finished.

Usage would look something like this:

```scala
import org.scalatest.fixture.ResourceManagerFixture
import org.scalatest.funsuite.FixtureAnyFunSuite

class DatabaseTest extends FixtureAnyFunSuite with ResourceManagerFixture {
  lazy val dbContainer = suiteScope {
    // create database container here
  }

  test("some database test") { use =>
    val dataSource = use {
      // create datasource here, with parameters obtained from `dbContainer`
    }
  }
}
```
The use of `lazy val` neatly avoids use of `var` and `null`. It also makes things efficient: if you never touch that `lazy val`, it won't be initialized at all. So you can e. g. define a database container in your base test suite class, and only those tests that interact with the database will actually spin up a container. This doesn't work with an `afterAll` method: if you define `lazy val someResource = …` in the suite and then call `someResource.close()` in `afterAll`, even if the `lazy val` wasn't used before, it will be initialized when `afterAll` runs only to then call `.close()` on it.
For the `Using.Manager` instance that is cleaned up after each test, I considered two approaches: pass it as a fixture parameter, or store it in a `var`. I ultimately went with a fixture because this allows tests to run in parallel.
This approach also works a lot better than loan-fixture methods. If you need multiple resources, the calls need to be nested, negatively impacting readability. With a `Using.Manager`, you just assign multiple vals:
```
  val resource1 = use(createResource())
  val resource2 = use(createOtherResource())
```
Suite-scoped and test-scoped resources can trivially be combined. And if any resource initialization fails, resources allocated before will be cleaned up correctly. You can even allocate resources in a loop!

I have used this approach to great effect in various projects and it really makes resource management a lot less painful than what's currently possible. So I wanted to share it with the wider ScalaTest community.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/fixture/ResourceManagerFixture.scala
+++ b/jvm/core/src/main/scala/org/scalatest/fixture/ResourceManagerFixture.scala
@@ -1,0 +1,66 @@
+/*
+ * Copyright 2001-2024 Artima, Inc.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.scalatest
+package fixture
+
+import scala.util.Using
+
+/**
+ * A Trait that facilitates the management of resources that need to be
+ * disposed of after using them.
+ * It provides two separate instances of the <code>Using.Manager</code>
+ * class from the Scala standard library (see the Scala standard library
+ * documentationon <code>scala.util.Using</code> for details on how it
+ * works).
+ * 
+ * The first instance can be obtained by calling the <code>suiteScoped</code>
+ * method. Its purpose is to clean up resources that are shared between
+ * different tests in a suite. This method can only be called after test
+ * execution has started. It is therefore recommended to use it to initialize
+ * a <code>lazy val</code> in the Suite class. This way, the 
+ * <code>lazy val</code> can be used to access the resource while deferring
+ * the <code>suiteScoped</code> call until test execution has started.
+ * 
+ * The second instance is passed as a fixture parameter to the test.
+ * It is cleaned up after the test has completed and can therefore be used
+ * to clean up resources that are only required inside a single test.
+ *
+*/
+trait ResourceManagerFixture extends FixtureTestSuite {
+  override type FixtureParam = Using.Manager
+
+  private var suiteManagerVar: Using.Manager = null
+
+  protected def suiteScoped = Option(suiteManagerVar).getOrElse {
+    throw new IllegalStateException("""
+      |`suiteScoped`` cannot be called from outside a test.
+      |Use a `lazy val` to store Suite-scoped resources in order to defer
+      |initialization of such resources until the start of a test.
+    """.stripMargin)
+  }
+
+  protected def withFixture(test: OneArgTest): Outcome =
+    Using.Manager { manager =>
+      withFixture(test.toNoArgTest(manager))
+    }.get
+
+  override protected def runTests(testName: Option[String], args: Args): Status =
+    Using.Manager { manager =>
+      suiteManagerVar = manager
+      super.runTests(testName, args)
+    }.get
+}


artimasites:
@cla-bot[bot] check

cheeseng:
@mberndt123 ResourceManagerFixture looks nice, thanks for sharing with us.  One problem is that in our main branch (for upcoming 3.3.0 release), we still need to support Scala 2.11 and 2.12, where Using.Manager does not seems to available in the standard scala library yet, I'll see if I can port over the Using.Manager code for 2.11 and 2.12 build.

mberndt123:
Hey @cheeseng,

I think that People who can live with a Scala version from 2017 can also live without `ResourceManagerFixture`. So I would suggest simply not including this feature in the Scala 2.12 and 2.11 releases. This is easily possible with version-specific source code directories in sbt.

If you feel you have to support this on 2.12 and 2.11, then I consider it essential that the 2.13 artifacts still use the standard library facilities in order to ensure seamless interoperability. When I write a helper method in my code that takes a stdlib `Using.Manager`, then I want it to work seamlessly within a ScalaTest context.

mberndt123:
I've thought about this a bit more and I think there's too much overlap with #2367, which also concerns itself with resource management. I'd like to think about this a bit more and then propose a more unified approach to resource management.

mberndt123:
I currently see two issues with this design:
1. it can't be combined with other Fixtures, like `ConfigMapFixture`
2. it isn't straight-forward to use for `Async*` suites.

Do you have any ideas on how to solve these?

cheeseng:
@mberndt123 Sorry just see your new messages, fyi I have been working on adjusting your PR: 

https://github.com/cheeseng/scalatest/tree/mberndt123-resource-manager-fixture

I don't have idea about your questions yet, but I'll think about them tomorrow with a fresh mind.

Cheers.

mberndt123:
Would it be possible to just define `org.scalactic.Using` as `val Using = scala.util.Using` when compiling for Scala 2.13 and later? This would make things work seamlessly when somebody defines a `Using.Releasable` instance for their own types.

cheeseng:
> Would it be possible to just define `org.scalactic.Using` as `val Using = scala.util.Using` when compiling for Scala 2.13 and later? This would make things work seamlessly when somebody defines a `Using.Releasable` instance for their own types.

fyi I have made `ResourceManagerFixture` a generated class, which when it is generated for Scala 2.13 and 3 it will use scala.util.Using directly, instead of org.scalactic.Using, only build for 2.11 and 2.12 will be using org.scalactic.Using.  Hopefully that should address your concern above, do let me know if you have further concern. :)

cheeseng:
@mberndt123 fyi today I managed to implement an async version, pushed to the same branch.  I had changed it to generate `Releasable` instead of `Using`, and modified `org.scalactic.Using` to allow our async code to use it.

mberndt123:
Yes, it's rather unfortunate that the standard library doesn't allow manual lifecycle management of `Using.Manager`. So we'll have to roll our own, unless they're prepared to make that public.

cheeseng:
@mberndt123 fyi I submitted the following PR against your PR branch: 

https://github.com/mberndt123/scalatest/pull/1

Cheers.

mberndt123:
By the way, I hope the code actually compiles with my changes. I can't check because for some reason scalatest doesn't compile at all on my machine, I get dozens and dozens of errors (mostly around `TableAsserting` ) in code I've never touched 🤷🏻‍♂️

cheeseng:
@mberndt123 I submitted the following PR to fix the build in test: 

https://github.com/mberndt123/scalatest/pull/2

You should be able to run the build and test with: 

```bash
> sbt clean test
```

Make your heap size for the process to at least 10GB.

mberndt123:
Cool, thanks, it compiles now and `{,Async}ResourceManager{,Fixture}Spec` runs. So, anything else needed to merge this? Afaics we're good to go!

mberndt123:
I've made a PR to add some docs on the website:
https://github.com/scalatest/scalatest-website/pull/246

cheeseng:
@mberndt123 I just let it ran the CI build and all of them passed!  We'll need @bvenners 's approval before I can merge into `main` branch, and expect to be released as the upcoming 3.3.0.

bvenners:
@mberndt123 Sorry I hadn't looked at it yet. Will look through it tonight.

mberndt123:
@bvenners
Did you get around to this?

bvenners:
> @bvenners Did you get around to this?

@mberndt123 Sorry, not yet. I'll work on it with Chee Seng his Monday morning.

Sorry for the delays and thanks for your patience.

Bill

mberndt123:
Hi @bvenners,

I just resolved some merge conflicts on this PR since it's been waiting for review for a while. When you have a moment, could you please take a look and (hopefully!) approve it? I’d love to get this merged soon so I don’t have to deal with any more conflicts down the line.

Thanks a lot!

bvenners:
@mberndt123 Sorry for the delay, and thanks for pinging. I wanted to make sure I understood it and haven't had time to study it yet, but will get on it very soon as we have releases coming up.

jvm/core/src/main/scala/org/scalatest/fixture/ResourceManagerFixture.scala
```diff
@@ -0,0 +1,71 @@
...
+ *     // Stream will be closed after the test
+ *   }
+ * }
+ * }}}
+ *
+ * @note This trait extends `FixtureTestSuite` and sets `FixtureParam` to `Using.Manager`.
+ * @see [[org.scalactic.Using.Manager]] for more details on resource safety and usage patterns.
+ */
+trait ResourceManagerFixture extends FixtureTestSuite {
+
```
bvenners, line 48: I am making a branch with this changed to use a self-type. I'll make a pr against your branch.

jvm/core/src/main/scala/org/scalatest/fixture/AsyncResourceManagerFixture.scala
```diff
@@ -0,0 +1,82 @@
...
+ *     }
+ *   }
+ * }
+ * }}}
+ *
+ * @note This trait extends `FixtureAsyncTestSuite` and sets `FixtureParam` to `Using.Manager`.
+ * @see [[org.scalactic.Using.Manager]] for details on how resource management is handled safely.
+ */
+trait AsyncResourceManagerFixture extends org.scalatest.FixtureAsyncTestSuite {
+  /**
```
bvenners, line 51: I also changed this to use a self type rather than extends. Will submit a PR.

bvenners:
@mberndt123 I was unable to make a PR against your branch in your fork, so I made a copy and make a PR against the copy: https://github.com/scalatest/scalatest/pull/2381. Please take a look.

cheeseng:
@bvenners I reran the CI build with your changes, and all are green, do you mind approve this PR so that I can merge it in?

Thanks.

bvenners (approved):

PR #2367, opened 2025-05-04
Add composable fixtures

mberndt123:
Hey there,

As I have remarked in #2366, fixture traits don't work as a unit of composition.
This is my attempt at an abstraction for fixtures that can be composed.

If you like this idea, it can easily also be applied to Async suites. For now, let's start with the sync case.


mberndt123:
I've thought about this a bit more and I think there's too much overlap with #2366, which also concerns itself with resource management. I'd like to think about this a bit more and then make another PR. Thanks.

PR #2368, opened 2025-05-06
FixtureAsyncTestSuite.withFixture consistent with FixtureTestSuite.withFixture

cheeseng:
Made withFixture method in FixtureAsyncTestSuite.scala protected, to be consistent with withFixture method in FixtureTestSuite.scala.

This is a breaking change, so let's do it only in the upcoming 3.3.0.


bvenners (approved):
It is better to be consistent, and this should not break much if any code in the wild.

PR #2381, opened 2025-07-09
Resource manager fixture edits

bvenners:
this is against a copy of https://github.com/scalatest/scalatest/pull/2366, because I was unable to make a PR against the real thing.


cheeseng:
Yes, closing as the changes are already in the main.

PR #2395, opened 2025-10-07
Update Expectations scaladoc

bvenners:
I also made a few tweaks to error messages in diagrams.


bvenners:
Oops, forgot to do scaladoc for the companion object. Will push that to this branch then un-draft the PR.

bvenners:
ok it is ready now.

PR #2398, opened 2025-10-08
Add support for passing initial seed to ScalaCheck properties

### LINKED ISSUE #2397: Setting seed via `JUnitRunner`
We're using Gradle to run our Scala tests and `JUnitRunner`. I combed through the code, but I couldn't find a way to get it to set the seed like you would normally do with the `-S` option with other tools. Is there a way to control the seed via a system property, or some way via `JUnitRunner`?

s-nel:
It's not always possible to configure the ScalaCheck seed via `-S <seed>` argument. For example when using the JUnit runner via Gradle (see #2397). This PR allows us to specify a seed at the property or spec level to help reproducing property failures. This will be used by a subsequent PR in https://github.com/s-nel/scalatestplus-scalacheck that will use this parameter and fall back to the global seed in `org.scalatest.prop.Seed`, or a random seed if the global seed is absent.

Fixes: https://github.com/scalatest/scalatest/issues/2397

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/prop/Configuration.scala
+++ b/jvm/core/src/main/scala/org/scalatest/prop/Configuration.scala
@@ -16,6 +16,8 @@
 package org.scalatest.prop
 
 import org.scalactic.anyvals.{PosInt, PosZDouble, PosZInt}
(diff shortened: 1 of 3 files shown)


artimasites:
@cla-bot[bot] check

PR #2401, opened 2025-10-16
Fix implicit conversion comment in ExampleSuite

nightscape:
The example as it stands has one line of code that evaluates to `Unit`, so no implicit conversion should be happening. Only when the return value is a `Fact` it would be converted to an `Assertion`.


artimasites:
@cla-bot[bot] check

PR #2425, opened 2026-05-28
Use Scala Native `test-interface-sbt-defs` instead of `test-interface`

zainab-ali:
See https://github.com/scala-native/scala-native/issues/4844

Testing libraries and frameworks should depend on Scala Native's `test-interface-sbt-defs`, not `test-interface`.
 - `test-interface-sbt-defs` defines a common interface for unit tests. It follows `early-semver` versioning.
 - `test-interface` is a server implementation that must match the version of the Scala Native plugin. It follows `strict` versioning.

This was picked up by SBT [1.12.10](https://eed3si9n.com/sbt-1.12.10), which added eviction checks for test dependencies.

## How to reproduce

Create an SBT project with the following:

```scala
// project/build.properties
sbt.version = 1.12.11
```

```scala
// project/plugins.sbt
addSbtPlugin("org.scala-native" % "sbt-scala-native" % "0.5.11")
```

```scala
// build.sbt
scalaVersion := "3.3.7"
enablePlugins(ScalaNativePlugin)

ThisBuild / evictionWarningOptions := (ThisBuild / evictionWarningOptions).value
  .withConfigurations(List(Compile, Test))

libraryDependencies += "org.scalatest" %%% "scalatest" % "3.2.20" % Test
```

The `sbt update` command gives:

```
[error] (update) found version conflict(s) in library dependencies; some are suspected to be binary incompatible:
[error]
[error] 	* org.scala-native:test-interface_native0.5_3:0.5.11 (strict) is selected over 0.5.10 for test
[error] 	    +- default:scala-native-seed-project_native0.5_3:0.1.0-SNAPSHOT (depends on 0.5.11)
[error] 	    +- org.scalatest:scalatest-core_native0.5_3:3.2.20    (depends on 0.5.10)
```

### DIFF
--- a/project/DottyBuild.scala
+++ b/project/DottyBuild.scala
@@ -373,7 +373,7 @@ trait DottyBuild { this: BuildCommons =>
                                        |import org.scalactic._
                                        |import Matchers._""".stripMargin,
       libraryDependencies += "org.scala-lang.modules" %%% "scala-xml" % "2.4.0",
-      libraryDependencies += ("org.scala-native" %%% "test-interface" % nativeVersion),
+      libraryDependencies += ("org.scala-native" %%% "test-interface-sbt-defs" % nativeVersion),
       packageManagedSources,
       Compile / sourceGenerators += Def.task {
         GenScalaTestDotty.genScalaNative((Compile / sourceManaged).value / "scala", version.value, scalaVersion.value) ++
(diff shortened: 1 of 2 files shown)


artimasites:
@cla-bot[bot] check

cheeseng:
@zainab-ali I tried to reproduce the problem in branch https://github.com/scalatest/scalatest/pull/2428 by upgrading the sbt version to 1.12.12, but unable to reproduce the mentioned error, I wonder if I should include the following in your example into the build to reproduce: 

```
ThisBuild / evictionWarningOptions := (ThisBuild / evictionWarningOptions).value
  .withConfigurations(List(Compile, Test))
```

I am not sure what it is doing, mind to help explaining?

Thanks.

zainab-ali:
Thanks for looking into this! To reproduce you'll need to set the following options:

```
ThisBuild / evictionWarningOptions := (ThisBuild / evictionWarningOptions).value
  .withConfigurations(List(Compile, Test))
```

Note that these are a default in SBT 2.0.

This applies the SBT eviction check to both `Compile` and `Test` configurations. In SBT `1.12.10` and below, the eviction check only applied to `Compile` dependencies. This meant that binary incompatibilities in test dependencies would manifest as runtime errors. 

SBT `1.12.11` attempted to roll out the eviction check to test dependencies by making the setting above a default. However, this caused almost all Scala Native builds to fail due to testing libraries depending on `test-interface` instead of `test-interface-sbt-defs`. To work around this, the `evictionWarningOptions` default configurations were reverted back to `Compile` in SBT `1.12.12`.

The linked Scala Native issue has more context: https://github.com/scala-native/scala-native/issues/4844.

bvenners (approved):

PR #2440, opened 2026-08-31
Preserve shrink structure through edges — carry List[RoseTree[T]] in Generator.initEdges/next

cheeseng:
## Summary
Refactor Generator.initEdges and Generator.next to carry List[RoseTree[T]] edges instead of bare List[T], so that composite generators — most notably the Option generator — preserve their shrink structure through edges.
Previously, edge handling only recorded the head value of each edge (List[T]), discarding the underlying RoseTree. This meant that when an edge was generated as part of a composed generator (e.g. a Some(...) inside an Option), its associated shrinks were lost once the edge flowed through the composition. By threading the full RoseTree[T] along each edge, composed generators now shrink edges correctly, matching the behavior of next.

## What changed
- Generator.initEdges and Generator.next now return/carry List[RoseTree[T]] for edges rather than List[T].
- Updated all generator implementations that override or consume these methods (in Generator.scala, CommonGenerators.scala, and PropCheckerAsserting.scala) to produce and propagate the richer edge type.
- Regenerated the GeneratorFor* templates (project/GenGen.scala) to reflect the new edge type.
- Added/updated edge tests in CommonGeneratorsSpec.scala, GeneratorSpec.scala, and HavingLengthsBetweenSpec.scala, including a regression test confirming that an edge value now shrinks to its canonical form (e.g. 2147483647 → 3).
- Fixes an intermittent shrink bug where an edge's shrink chain was broken after composition.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/enablers/PropCheckerAsserting.scala
+++ b/jvm/core/src/main/scala/org/scalatest/enablers/PropCheckerAsserting.scala
@@ -187,7 +187,7 @@ object PropCheckerAsserting extends ExpectationPropCheckerAsserting {
       val maxSize = PosZInt.ensuringValid(minSize + config.sizeRange)
 
       @tailrec
-      def loop(succeededCount: Int, discardedCount: Int, edges: List[A], rnd: Randomizer, initialSizes: List[PosZInt], initSeed: Long): PropertyCheckResult = {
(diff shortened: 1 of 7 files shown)


bvenners (approved):
@cheeseng Please adjudicate the copilot review before merging.

PR #2441, opened 2026-09-01
Made classify mimic forAll's sizing behavior.

cheeseng:
Changes:
1. Added minSize (default PosZInt(0)) and sizeRange (default PosZInt(100)) params to classify — matching Configuration defaults, preserving all existing callers and enabling configurability.
2. Replaced the hardcoded SizeParam(0, 100, 100) with forAll's sizing: precompute a size sequence via a local calcSizes(minSize, maxSize, rnd) (10 sorted sizes seeded with minSize), consume them in order, then fall back to choosePosZInt(minSize, maxSize), passing SizeParam(0, maxSize, size) each iteration.
3. Edges now use count / 5 (like forAll's minSuccessful / 5) instead of a fixed 100, per the "same edges approach" note.
4. Removed the // classify will need to use the same sizing algo as forAll and // TODO: I think this need to mimic forAll comments.

### DIFF
--- a/jvm/core/src/main/scala/org/scalatest/prop/CommonGenerators.scala
+++ b/jvm/core/src/main/scala/org/scalatest/prop/CommonGenerators.scala
@@ -2347,34 +2347,60 @@ trait CommonGenerators {
     *
     * @param count the number of values to generate
     * @param genOfA the [[Generator]] to use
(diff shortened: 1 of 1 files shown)


bvenners (approved):
Please adjudicate the Copilot comment before merging.

